PR #30, opened 2021-10-28
Add support for SSL, remove dependency from vdms

luisremis:


aperturedb/Connector.py
```diff
@@ -1,37 +1,182 @@
-import os
+#
+# The MIT License
+#
+# @copyright Copyright (c) 2017 Intel Corporation
+# @copyright Copyright (c) 2021 ApertureData Inc
```
vishakha041, line 5: I wonder if we can do this and if it means anything

vishakha041 (dismissed):
As far as I can tell, seems fine.

andreim-sv (dismissed):
Good to go.

PR #31, opened 2021-10-29
Add file sources

luisremis:


aperturedb/ImageLoader.py
```diff
@@ -86,11 +118,50 @@ def load_image(self, filename):
...
+        return False, None
+
+    def load_s3_url(self, s3_url):
+        s3 = boto3.client('s3')
+
+        try:
+            filename = "/tmp/" + s3_url.split("/")[-1]
+            bucket_name = s3_url.split("/")[2]
+            object_name = s3_url.split("s3://" + bucket_name + "/")[-1]
+            s3.download_file(bucket_name, object_name, filename)
```
luisremis, line 151: it seems you can avoid downloading to a file here as well:
https://stackoverflow.com/questions/37087203/retrieve-s3-file-as-object-instead-of-downloading-to-absolute-system-path

aperturedb/ImageLoader.py
```diff
@@ -86,11 +118,50 @@ def load_image(self, filename):
...
 
+    def load_url(self, url):
+
+        retries = 0
+        while True:
+            imgdata = requests.get(url)
+            if imgdata.ok:
+                if self.check_image:
+                    filename = "/tmp/" + url.split("/")[-1]
+                    fd = open(filename, "wb")
```
luisremis, line 129: you can avoid writing to a file and checking the image from buffer directly. Here is an example:
https://github.com/luisremis/visual_storm/blob/master/yfcc100m/python/eval/mysql_eval/MySQLQuery.py#L212

test/run_test.sh
```diff
@@ -28,6 +28,9 @@ echo "Generating input files..."
 python3 generateInput.py
 echo "Done generating input files."
 
+# Needs the boto package for s3 test
+pip3 install boto3
```
luisremis, line 32: this can go away, should not be here.

test/test_S3Loader.py
```diff
@@ -0,0 +1,68 @@
...
+import time
+import unittest
+
+import dbinfo
+import numpy  as np
+
+from aperturedb import Connector, Status
+from aperturedb import ImageLoader
+
+class TestLoader(unittest.TestCase):
```
luisremis, line 11: this is the same as test_Loader.py. It would be better to simply import test_Loader.py here.
Or, you can just move the class TestS3Loader to test_Loader.py. S3 loader is yet another type of loader.

luisremis (commented):
good addition

andreim-sv (dismissed):
Good to go.

aperturedb/ImageLoader.py
```diff
@@ -83,14 +111,75 @@ def load_image(self, filename):
             return True, buff
         except:
             print("IMAGE ERROR:", filename)
+        return False, None
+
+    def check_image_buffer(self, img):
+        try:
+            # Warning -> cv2.imdecode returns None for some images
```
luisremis, line 118: this comment is from a copy-paste, does not apply here.

luisremis (commented):
looking great! will remove that copy-paste comment and merge.

PR #39, opened 2022-02-21
try_catch_user_input

andreim-sv:
### Purpose
Report user errors when unexpected or incorrect keys are used in Loader.

### Detail
When Generator uses keys other than the ones expected by the Loader (e.g. `ref1_value` instead of `ref1_val`), the author of the Generator would get a detailed error message.


luisremis (approved):
great!

PR #44, opened 2022-02-28
Draft API specification.

gsaluja9:
**Starting to implement an Object mapper API.**
It would enable the users to interact with athena using an API that alleviates the need to be familiar with the [native API](https://docs.aperturedata.io/index.html).

It is going to be designed on the guidelines of a similar ORM library, for example [SQL Alchemy](https://www.sqlalchemy.org/) or [Django's ORM](https://docs.djangoproject.com/en/4.0/topics/db/queries/).

The following are some considerations for implementing it this time:
- It will support CRUD operations for the object types covered by the native API.
- It will treat the Entities in a more or less common way, and implement the basic DB operations (rather than the prescribed ways of actions on these objects.
- The main goal is will be to add simplicity and intuitiveness for interacting with VDMS. The current API requires the users to familiarize themselves with the API.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,3 +132,7 @@ dmypy.json
 *.adb.csv
 *.jpg
 *.npy
+
+env3/*
+test/aperturedb/db/*
+test/input/blobs/*
--- a/README.md
+++ b/README.md
@@ -13,3 +13,46 @@ elements in ApertureDB at the object level.
 * NotebookHelpers.py provides helpers to show images/bounding boxes on Jupyter Notebooks
 
 For more information, visit https://aperturedata.io
+
+
+## Object Mapper API
+This API allows the user to interact with the database with the idea of objects that can be Created, Read, Updated and Deleted through an intutive interface involving these objects.
+
+It supports the following types of Objects in the Database as Python objects:
+| Object | Dcumentation |
+| ----- | -----|
+| Blob||
+| BoundingBox || 
+| Descriptor||
+| DescriptorSet||
+| Entity||
+| Image||
+| Video||
+
+
+### Examples
+**Create an Image**
+
+```
+from aperturedb import Connector, Images
+images = Images.Images(db_connector)
+with open(abs_path, "rb") as instream:
+    img_data = instream.read()
+    images.create_new(
+        properties: {
+            description: "A new image API",
+            custom_index: 92
+        }
+    )
+    image = image.save()
+    # image is an object with _uniqueid and arbitrary properties, and methods.
+```
+
+**Retrieve an Image**
+
+```
+from aperturedb import Connector, Images
+images = Images.Images(db_connector)
+
+image = image.get(id=92)
+```
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ It supports the following types of Objects in the Database as Python objects:
 
 ```
 from aperturedb import Connector, Images
+
+db_connector = Connector.Connector(user="admin", password="admin")
 images = Images.Images(db_connector)
 with open(abs_path, "rb") as instream:
     img_data = instream.read()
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -132,3 +132,7 @@ dmypy.json
 *.adb.csv
 *.jpg
 *.npy
+
+env3/*
+test/aperturedb/db/*
+test/input/blobs/*
```
luisremis, line 138:
```suggestion
test/input/blobs/*

```

README.md
```diff
@@ -13,3 +13,65 @@ elements in ApertureDB at the object level.
 * NotebookHelpers.py provides helpers to show images/bounding boxes on Jupyter Notebooks
 
 For more information, visit https://aperturedata.io
+
+
+## Object Mapper API
+This API allows the user to interact with the database with the idea of objects that can be Created, Read, Updated and Deleted through an intutive interface involving these objects.
```
luisremis, line 19:
```suggestion
This API allows the user to interact with the database with the idea of objects that can be 
Created, Read, Updated and Deleted through an intuitive interface involving these objects.
```

README.md
```diff
@@ -13,3 +13,65 @@ elements in ApertureDB at the object level.
...
+
+It supports the following types of Objects in the Database as Python objects:
+| Object | Dcumentation |
```
luisremis, line 22:
```suggestion
| Object | Documentation |
```

README.md
```diff
@@ -13,3 +13,65 @@ elements in ApertureDB at the object level.
...
+| BoundingBox || 
+| Descriptor||
+| DescriptorSet||
+| Entity||
+| Image||
+| Video||
+
+
+### Examples
+**Create an Image**
```
luisremis, line 34: this is a great start, but what we would like is two separate sets of documentation: 
1) docstrings or similar, to let users understand the API. Ideally, we would like to generate an html documentation (like this: https://pytorch.org/docs/stable/torch.html (which actually uses read-the-docs))
2) Example documentation, for users to see some use-cases. For the examples, you can consider using read-the-docs/sphinx as we are using for documentation ApertureDB API

README.md
```diff
@@ -13,3 +13,65 @@ elements in ApertureDB at the object level.
...
+
+```
+from aperturedb import Connector, Images
+
+db_connector = Connector.Connector(user="admin", password="admin")
+images = Images.Images(db_connector)
+with open(abs_path, "rb") as instream:
+    img_data = instream.read()
```
luisremis, line 42: where is img_data used?

README.md
```diff
@@ -13,3 +13,65 @@ elements in ApertureDB at the object level.
...
+```
+
+**Connect an Image to BoundingBox**
+```
+from aperturedb import Connector, Images, BoundingBoxes
+
+db_connector = Connector.Connector(user="admin", password="admin")
+
+bboxes = BoundingBoxes.BoundingBoxes(db_connector)
+bbox = bboxes.get(id=34)
```
luisremis, line 71: this is a good start. Are there other, more friendly, ways of interacting with the objects without having to specify ids? I would try to come up with something higher level than that.

aperturedb/BoundingBoxes.py
```diff
@@ -0,0 +1,7 @@
+from aperturedb.Repository import Repository
+
+
+class BoundingBoxes(Repository):
+    def __init__(self, db, batch_size=100):
+        super().__init__(db)
+        self._object_type = "BoundingBox"
```
luisremis, line 7:
```suggestion
        self._object_type = "BoundingBox"

```

test/test_Image.py
```diff
@@ -0,0 +1,107 @@
...
+    def test_enumerate(self):
+        for image in self.images.filter():
+            self.assertIsNotNone(image)
+            self.assertTrue(hasattr(image, "_uniqueid"))
+
+        
+
+        
+if __name__ == '__main__':
+    unittest.main()
```
luisremis, line 107:
```suggestion
    unittest.main()

```

test/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    images = Images.Images(db)
+    entities = Repository.Repository(db)
+    bboxes = BoundingBoxes(db)
+    print("Cleaning up DB")
+    for image in images.filter():
+        image.delete()
+    for entity in entities.filter():
+        entity.delete()
+    for bbox in bboxes.filter():
+        bbox.delete()
```
luisremis, line 19:
```suggestion
        bbox.delete()

```

test/utils.py
```diff
@@ -0,0 +1,19 @@
+from aperturedb import Images, Repository
+from aperturedb.BoundingBoxes import BoundingBoxes
+
+
+    
+def cleanDB(db):
```
luisremis, line 6: scary name for a method.

luisremis (changes requested):
This is going in the right direction. Before reviewing any implementation details, it would be nice to have a more complete description of the API, where most of the discussion will go.

gsaluja9:
done in another.

PR #45, opened 2022-03-15
Feature/api docs 40

### LINKED ISSUE #40: Create Documentation through docstring for the package
We currently have very poor documentation on the APIs for our Python Package. 

This task can be splitted in 2:
1) Add a mechanism to automatically generate html, user-facing, documentation from docstrings
2) Add documentation for all the modules in the package. Specially attention to the interface methods that are user-facing.

### LINKED ISSUE #43: Enhance support for Images at OM level
We should enable users to insert images using our OM API. 

The APIs for inserting images will be the template for the rest of objects in our API.

gsaluja9:
**This is a skeleton of the OM API.**

The command to auto generate the docs using sphinx-apidoc locally is `cd docs && bash gen_docs.sh`. Please be aware that this is better run off a virtual env, as it installs certain packages, which have not been fine tuned.

Alternatively, this generated documentation has been exposed from a light sail instance as:
http://18.207.180.191:8000/index.html

All the newer classes have been put in OM package.

There are certain tests too, but those won't work with CI yet, till the cleanDB is merged into the code base.

The initial interface is based on the following:
Image from iOS

Fix #43 
Fix #40


docs/Makefile
```diff
@@ -0,0 +1,20 @@
...
+# You can set these variables from the command line, and also
+# from the environment for the first two.
+SPHINXOPTS    ?=
+SPHINXBUILD   ?= sphinx-build
+SOURCEDIR     = .
+BUILDDIR      = _build
+
+# Put it first so that "make" without argument is like "make help".
+help:
+	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
andreim-sv, line 13: Just curious what `$(0)` would expand to?

docs/gen_docs.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
andreim-sv, line 2: Would you consider saying `set -e` here?

requirements.txt
```diff
@@ -0,0 +1,121 @@
...
+tornado==6.1
+traitlets==5.1.1
+typing-extensions==4.1.1
+Unidecode==1.3.4
+urllib3==1.26.8
+vdms==0.0.16
+wcwidth==0.2.5
+webencodings==0.5.1
+wrapt==1.14.0
+zipp==3.7.0
```
andreim-sv, line 121: Just curious: do the versions have to be specific, i.e. `==`, rather than `>=`?

test/test_om_image.py
```diff
@@ -0,0 +1,106 @@
...
+                    "timestamp": str(datetime.now()),
+                    "path": path
+                }, 
+                operations=[], 
+                blob=img_data)
+    
+    def test_create(self):
+        for i in range(10):
+            file_path = f"input/images/000{i}.jpg"
+            abs_path  = os.path.join(os.path.join( os.path.dirname( __file__) , file_path))
```
andreim-sv, line 51: Would `os.path.abspath(file_path)` work here?

andreim-sv (approved):
Looks good, it was a pleasure to read.

.gitignore
```diff
@@ -132,3 +132,6 @@ dmypy.json
 *.adb.csv
 *.jpg
 *.npy
+env3/
+test/aperturedb/db/
+test/input/blobs/
```
luisremis, line 137:
```suggestion
test/input/blobs/

```

.vscode/settings.json
```diff
@@ -0,0 +1,30 @@
...
+            }
+        ]
+    },
+    "python.testing.pytestArgs": [
+        "test",
+        "--capture=tee-sys"
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true
+}
```
luisremis, line 30:
```suggestion
}

```

aperturedb/OM/Entity.py
```diff
@@ -0,0 +1,74 @@
+from typing import ByteString, List, Optional, Dict
+from aperturedb import Connector
+
+
+class Entity:
```
luisremis, line 5: this is a confusing name, because we have one type of object on ApertureDB that is called "Entity"

aperturedb/OM/Entity.py
```diff
@@ -0,0 +1,74 @@
...
+
+        if self._blob is not None:
+            resp, _ = self._db.query(query, [self._blob])
+        else:
+            resp, _ = self._db.query(query)
+
+        for property, value in resp[0][self._add_command()].items():
+            setattr(self, property, value)
+        
+        return self
```
luisremis, line 74:
```suggestion
        return self

```

aperturedb/OM/Repository.py
```diff
@@ -0,0 +1,169 @@
...
+        self._object_type = "Entity"
+        self._db = db
+        self._entity_types = {
+            "Entity": Entity
+        }
+
+    def _find_command(self):
+        return f"Find{self._object_type}"
+
+    def get(self, id) -> Entity:
```
luisremis, line 31: we should try to avoid exposing uniqueid to the user, given that the OM is a higher level abstraction.
but the id in this case can be a local id (or the index) of the object within the repository.

luisremis (changes requested):
generated docs look pretty good! it would be nice to discuss some aperturedb notions we want to represent on the OM and do another session of brainstorming.

docs/gen_docs.sh
gsaluja9: will do.

aperturedb/OM/Entity.py
gsaluja9: Yes, I was thinking the same. I am trying to think alternative, Object and unit comes to mind. I think will rename it to Unit. Do you have any specific word in mind?

docs/Makefile
gsaluja9: Ha. this was auto generated, but it'd be good to know.
I got this : https://www.gnu.org/software/make/manual/html_node/Call-Function.html

requirements.txt
gsaluja9: I am sure the versions won't even matter. I'll reduce this to what I specifically installed to have this running rather than the output of pip freeze.

aperturedb/OM/Entity.py
andreim-sv: "Thingy", "Thing", "Sometning", "ADObject", "Substance", "Root", "Stuff".

JoshStoddard: FWIW, I vote for something like `Object` or `BaseObject` since this is the Object Mapper.

.vscode/settings.json
```diff
@@ -0,0 +1,30 @@
+{
```
JoshStoddard, line 1: maybe `.gitignore` `.vscode/`?

docs/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
JoshStoddard, line 1: remove this windows-specific file?

PR #46, opened 2022-03-17
More to status

luisremis:


aperturedb/ParallelLoader.py
```diff
@@ -14,15 +14,14 @@ class ParallelLoader:
 
     def __init__(self, db, dry_run=False):
 
-        self.db = db
+        self.db = db.create_new_connection()
```
gsaluja9, line 17: Is there something that the master will do with a new connection ? The worker will create a new connection in any case, right?
Perhaps I could use a primer on our connection related changes :)

aperturedb/ParallelLoader.py
```diff
@@ -118,19 +119,24 @@ def print_stats(self):
...
         print("Total queries executed:", total_queries_exec)
-        print("Avg Query time (s):", np.mean(times))
-        print("Query time std:", np.std (times))
-        print("Avg Query Throughput (q/s)):",
-                1 / np.mean(times) * self.numthreads)
 
-        msg = "(" + self.type + "/s):"
-        print("Overall insertion throughput", msg,
-                self.total_elements / self.ingestion_time)
+        if total_queries_exec == 0:
```
gsaluja9, line 123: is this to prevent division by 0 on l139? It might still be of use to have some times if they get reported, though I have not run this enough times to see how those counters fill up.

aperturedb/Status.py
```diff
@@ -132,6 +136,50 @@ def count_connections(self, connections_class, constraints={}):
             total_connections = res[0]["FindConnection"]["count"]
         except:
             total_connections = 0
-            print(self.connector.get_last_response_str())
+            self.connector.print_last_response()
 
         return total_connections
+
+    def count_descriptorsets(self):
```
gsaluja9, line 143: Is this to be removed ? Not invoked from anywhere.

aperturedb/Status.py
```diff
@@ -132,6 +136,50 @@ def count_connections(self, connections_class, constraints={}):
...
+        try:
+            res, blobs = self.connector.query(q)
+            total_descriptor_sets = res[0]["FindDescriptorSet"]["count"]
+        except:
+            total_descriptor_sets = 0
+            self.connector.print_last_response()
+
+        return total_descriptor_sets
+
+    def remove_descriptorset(self, set_name):
```
gsaluja9, line 162: Same as above .

gsaluja9 (changes requested):

aperturedb/ParallelLoader.py
luisremis: I simply create a new connection to test, at construction time, that the connection will work. Otherwise, it will fail and throw.

aperturedb/ParallelLoader.py
luisremis: yes, it is mainly to prevent division by zero. When no queries run, it is because there is something off with the input data (an exception is usually thrown). There is no need to report any other metric if no query succeeded.

aperturedb/Status.py
luisremis: these are used by clients. I added because I use it quite often. I think we should rename Status altogether to "Utils" or something of the kind. Status is more than just showing the status now. We can have Utils that have many utility methods, together with showing the status, what do you think?

gsaluja9: This can go into Utils, or even a class like Images for [DescriptorSets](https://docs.aperturedata.io/commands/AddDescriptorSet.html) OM .

luisremis: cool sounds good, let me do some refactoring then, and I need to make sure those tests pass!

test/test_Base.py
```diff
@@ -0,0 +1,46 @@
...
+
+                attempts += 1
+                time.sleep(1) # sleeps 1 second
+
+            if attempts > 10:
+                print("Failed to connect to ApertureDB after 10 attempts")
+                exit()
+
+    def create_connection(self):
+        return Connector.Connector(self.db_host, self.db_port, self.user, self.password)
```
luisremis, line 46:
```suggestion
        return Connector.Connector(self.db_host, self.db_port, self.user, self.password)
 
```

aperturedb/Status.py
luisremis: done! added more stuff to Utils as well.

gsaluja9 (approved):
Looks good.

PR #49, opened 2022-03-23
Feature/existing docs

gsaluja9:
Extracted the changes related to documentation for the existing classes so that it may be merged independently from the OM API related stuff.

Also ran a lint tool 'black' to make the code style consistent and remove the trail white space, etc.

TODO:

1. Run the CI part of this sphinx-docs and host it somewhere (internal).
2. Proofread the rendered documentation. Formatting might have had unintended side effects.
3. Fix the errors and publish it external.


aperturedb/BBoxLoader.py
```diff
@@ -1,63 +1,69 @@
 from aperturedb import ParallelLoader
 from aperturedb import CSVParser
 
-HEADER_X_POS  = "x_pos"
-HEADER_Y_POS  = "y_pos"
-HEADER_WIDTH  = "width"
+HEADER_X_POS = "x_pos"
```
luisremis, line 4: what happened here? it used to look so nice all with equal indentation, easier to read. Maybe so automatic rule from your editor?

docs/conf.py
```diff
@@ -0,0 +1,91 @@
...
+napoleon_use_admonition_for_notes = False
+napoleon_use_admonition_for_references = False
+napoleon_use_ivar = False
+napoleon_use_param = True
+napoleon_use_rtype = True
+napoleon_preprocess_types = False
+napoleon_type_aliases = None
+napoleon_attr_annotations = True
+
+add_module_names = False
```
luisremis, line 91:
```suggestion
add_module_names = False

```

docs/gen_docs.sh
```diff
@@ -0,0 +1,12 @@
...
+cd ..
+#Install the dependencies
+pip install -r requirements.txt
+
+#Install aperturedb into site packages.
+python setup.py install
+cd -
+
+#Generate the docs from the docstrings.
+make clean && make html
```
luisremis, line 12:
```suggestion
make clean && make html

```

aperturedb/queryMessage_pb2.py
```diff
@@ -3,81 +3,115 @@
 # source: queryMessage.proto
 
 import sys
```
luisremis, line 5: this file is automatically generated. We should not modify it.

luisremis (commented):
looking great. We should generate the documentation through a docker image, and have another docker image that has a webserver + documentation to integrate with CI. Then everything will be automatically deployed. What a time to be alive. 

We also use read the docs and sphinx on aperturedb, so you should be able to copy and edit all of that from here: https://github.com/aperture-data/athena/tree/develop/docs/docker

Last step will be adding all the building of the docs and of the image with a webserver to actions, similarty to what we do here: https://github.com/aperture-data/aperturedb-python/blob/feature/existing_docs/.github/workflows/main.yaml#L53

aperturedb/BBoxLoader.py
gsaluja9: Not an editor, I used this https://black.readthedocs.io/en/stable/index.html . This seems like quite a mature and commonly used tool in OSS python projects. It has plugins for tons of editors, but it also integrates with github actions and commit prehook.

It is opinionated ...  yes, but so is python and pydoc standards.

So what this would offer is hassle from manually conforming to a consistent coding style.
We might not see it's immediate benefit as the number of people working with the repo is less, but you can immediately see there is one less thing to worry about as in the comment from docs/conf.py . I had not run it on root, but at aperturedb level. 

This makes even more sense as the repo is public, and it's hard to align to a common standard.
This tool just automates the process, without consuming any mental energy.

aperturedb/queryMessage_pb2.py
gsaluja9: agreed. will revert

aperturedb/BBoxLoader.py
luisremis: I love the idea of start using a style checker/enforcer for our Python projects. 

It would be great to enforce PEP8. Black is PEP8 compliant. But I think Black would force many style changes on top of PEP8 that are arbitrary rather than functional. Another thing is that Black is not very configurable, so it would mean using their style or die.

PEP8 is less restrictive and oriented towards readability. That would make it a great code style to enforce.
Here is a package that does that: https://pypi.org/project/pep8/
We can add a check on our CI and enforce it.

Here is another one, recommended by the very same Guido van Rossum: https://github.com/PyCQA/flake8

gsaluja9: Flake8 does look like a promising option. Let me try to change apply auto formatting using that.

test/test_torch_connector.py
```diff
@@ -143,13 +148,15 @@ def test_datasetWithMultiprocessing(self):
...
+        print("Distributed Data Loader Sequential Throughput (imgs/s):",
+              len(dataset) / (time.time() - start))
 
         # === Distributed Data Loader Shuffler
 
         # This will generate a random sampler, which will make the use
         # of batching wasteful
-        sampler     = torch.utils.data.DistributedSampler(dataset, shuffle=True)
+        sampler     = torch.utils.data.DistributedSampler(
+            dataset, shuffle=True)
```
luisremis, line 159:
```suggestion
        sampler = torch.utils.data.DistributedSampler(dataset, shuffle=True)
```

docs/index.rst
```diff
@@ -0,0 +1,20 @@
+.. aperturedb-python documentation master file, created by
```
luisremis, line 1: is this file automatically generated? If yes, we should not commit it.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
```
luisremis, line 1: mmmm interesting. How does this work?

luisremis (commented):
looking great!

docs/index.rst
gsaluja9: We modified it from the one which was generated as a default.

.pre-commit-config.yaml
gsaluja9: I installed pre-commit https://pre-commit.com/https://pre-commit.com/, and ran it's baseline command which did all the auto formatting. It put a hook into my .git/hooks as per the guide. Wanted to try how this works for multiple users, but it should get triggered by any non compliant commit.

test/test_torch_connector.py
gsaluja9: that was the auto formatter.

docs/docker/push_images.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# This script pushes the docker images tp dockerhub
```
luisremis, line 3:
```suggestion
# This script pushes the docker images to dockerhub
```

aperturedb/ConnectionLoader.py
```diff
@@ -5,28 +5,32 @@
...
+    .. note::
         Expects a csv file with the following columns:
 
-            ConnectionClass,Class1@PROP_NAME, Class2@PROP_NAME ... PROP_NAME_N,constraint_PROP1
+            ``ConnectionClass``, ``Class1@PROP_NAME``, ``Class2@PROP_NAME`` ... ``PROP_NAME_N``, ``constraint_PROP1``
+
+    Example csv file::
 
-        Example csv file:
         ConnectionClass,VD:IMG@id,Person@id,confidence,id,constaint_id
```
luisremis, line 19:
```suggestion
        ConnectionClass,_Image@id,Person@id,confidence,id,constaint_id
```

.github/workflows/main.yaml
```diff
@@ -54,3 +54,26 @@ jobs:
...
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PASS }}
+
+      - name: Build and push docker
+        run: |
+          cd docs/docker
+          bash build.sh ${GITHUB_REF##*/}
```
luisremis, line 78: this will build and tag based on the name of the branch. We should tag based on the version, which can be taken from the __init__.py file.

gsaluja9: If we use the same version for the release and these PRs won't the docker images get clobbered ? Or how do we pick the one intended for public consumption ?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,75 @@
...
+        run: |
+          pip3 install .
+          pip3 install ipython torch torchvision boto3
+          cd test
+          bash run_test.sh
+
+      - name: Build Notebook Docker
+        run: |
+          cd docker
+          bash build_images.sh
```
luisremis, line 52:
```suggestion
          bash build_images.sh

```

luisremis (dismissed):
looks good to go! Let's give it a try and refine!

luisremis:
I think the only missing part would be checking for compliance with pep8 on CI. We can add that later. Let's start merging and trying.

gsaluja9:
> I think the only missing part would be checking for compliance with pep8 on CI. We can add that later. Let's start merging and trying.

yea would like to experiment with a few options for that, before finalizing.

.github/workflows/release.yaml
gsaluja9: done

PR #51, opened 2022-04-01
Parallel query

luisremis:


aperturedb/ParallelLoader.py
```diff
@@ -2,117 +2,23 @@
 import time
```
gsaluja9, line 2: This class seems redundant as *Loader translates to essentially parallel queries .
Is there a use of parallel loader which you foresee ?

We should delete this and derive the loaders from parallelquery.

gsaluja9, line 2: This class seems redundant as *Loader translates to essentially parallel queries .
Is there a use of parallel loader which you foresee ?

We should delete this and derive the loaders from parallelquery.

aperturedb/ParallelQuery.py
```diff
@@ -0,0 +1,159 @@
...
+        self.loader_filename = "progress.log"
+
+    def  get_times(self):
+
+        return self.times_arr
+
+    def generate_batch(self, data):
+
+        """
+            Here we flatten the individual queries to run them as
```
gsaluja9, line 45: Are there other queries other than the Add stuff in loaders that would be using this ? All the loaders implement their own generate batch. Maybe an example in the docstring would explain the use case clearly.

gsaluja9, line 45: Are there other queries other than the Add stuff in loaders that would be using this ? All the loaders implement their own generate batch. Maybe an example in the docstring would explain the use case clearly.

aperturedb/ParallelLoader.py
luisremis: there is common functionality for all the loaders (as of now, just print_stats). Otherwise, we would have to implement a specific print_stat function for every loader.

aperturedb/ParallelQuery.py
luisremis: this is default behaviour of parallel queries: the generator is suppose to generate 1 query at the time, and we combine them to form a batch here. 
In the case of the loaders, we not only may combine queries, but also check some correctness dependent on the loader itself (for instace, the image loader checks that the query actually has the right parameters of AddImage and that there is a blob associated with it.)

Here is an example of a generator for a Find query:
https://github.com/aperture-data/athena-benchmark/pull/6/files#diff-8059b8cf2fdfd43fd87ce0d352f53c4dfa2deee54c4dfb058b1c0d142055c295R52

You are right, I should add more explanation to the docstring. 
I will add more comments

aperturedb/ImageDownloader.py
```diff
@@ -173,9 +173,9 @@ def print_stats(self):
         print("Throughput (images/s)):",
               1 / np.mean(times) * self.numthreads)
 
-        print("Total time(s):", self.ingestion_time)
+        print("Total time(s):", self.queries_time)
```
gsaluja9, line 176: This is reading odd, because there are no queries, and there is no Loader :),  Maybe we can define a function called get_aggregated_runtime() in ParallelLoader, and rename the class that Parallelizer. 

Also ParallelQuery sounds to be a specific use case of Parallelizer .
ParallelQuery should not be the base class

Let me know if we should discuss this offline.

gsaluja9 (changes requested):

aperturedb/ParallelQuery.py
gsaluja9: OK. I see. I think I updated a top level comment about inheritance . Let me know if that makes sense.

gsaluja9 (dismissed):
Looks good to go.

aperturedb/ImageDownloader.py
luisremis: Ok I gave it try. now we ended up with "batching" notion in Parallelizer, which is not used within that class, but needed in the logic for determining number of workers. I guess it is ok for now, given that the Parallelizer is already implemented.

andreim-sv (commented):
Looks good.

PR #54, opened 2022-04-04
Feature/more docs and index

gsaluja9:


docs/_static/parallelizer.dot
```diff
@@ -0,0 +1,19 @@
+digraph parallelizer {
```
gsaluja9, line 1: This is bases on a DSL : https://graphviz.org/doc/info/lang.html
Can be used to describe arbitrary graphs, that can be rendered.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v3
+      with:
+        python-version: '3.8.10'
+    - uses: pre-commit/action@v2.0.3
```
luisremis, line 16:
```suggestion
    - uses: pre-commit/action@v2.0.3

```

aperturedb/Connector.py
```diff
@@ -43,6 +43,22 @@
...
+    It lets the client execute any query as per the `Native Query specs <https://docs.aperturedata.io/>`_
+
+    Args:
+        str (host): Address of the host to connect to.
+        int (port): Port to connect to.
+        str (user): Username to specify while establishing a connection.
+        str (password): Password to specify while connecting to the db.
+        str (token): Token to use while connecting to the database.
+        object (session):
+        bool (use_ssl): Use SSL to encrypt communication with the database.
```
luisremis, line 60: is there a way to show the default values? In the case of use_ssl=True, for instance.

docs/_static/parallelizer.dot
luisremis: This is pretty cool, but you have to create it by hand? is there a way to automatically generate them? ()

docs/index.rst
```diff
@@ -3,8 +3,38 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-Welcome to aperturedb-python's documentation!
-=============================================
+Aperturedb-python's documentation.
```
luisremis, line 6:
```suggestion
ApertureDB Python Package Documentation.
```

luisremis (approved):
Looks good, thanks! take a look at minor suggestion as commits.

docs/_static/parallelizer.dot
gsaluja9: Will need to check something like that. Sounds quite radical, but I am not sure if there is something like that (cursory search)

aperturedb/Connector.py
luisremis: oh I see, great, this is good enough, sorry I missed it.

docs/_static/parallelizer.dot
luisremis: mmm cool, I guess we would have to add stuff to the docstring for any tool to autogenerate something like that. Probably not worth it at this point. But we can leave the ones you created for reference and over time add more and experiment. If we realize they go out of sync fast and hard to maintain, we just remove them all.

luisremis (approved):

PR #56, opened 2022-04-06
Add refresh token mechanism

luisremis:


gsaluja9:
this is odd. there should be 4 checks. Let me retrigger the workflow and see.

aperturedb/Connector.py
```diff
@@ -113,13 +115,51 @@ def _authenticate(self, user, password="", token=""):
...
+
         response, _ = self.query(query)
 
         self.session = response[0]["Authenticate"]
         if self.session["status"] != 0:
             raise Exception(self.session["info"])
 
+        self.authenticate_timestamp = time.time()
         self.session_token = self.session["session_token"]
+        self.refresh_token = self.session["refresh_token"]
```
gsaluja9, line 127: this does not seem to be used in the code, as it's always looked up from self.session .

gsaluja9 (approved):
looks good.

PR #57, opened 2022-04-06
Add refresh token mechanism

gsaluja9:
Cherry picked the commit onto a branch based on latest develop


gsaluja9:
This can be updated with pre-commit which can be installed with (as a one time) `pip install pre-commit autopep8 && pre-commit install`
Do a git commit --amend after that to trigger it in your local env.

PR #72, opened 2022-04-21
Formal documentation and signatures for aperturedb-python API.

gsaluja9:
Used the images methods as a reference point with some extra features.
This API allows people interact with a slice of data in ADB, put initial data, and also retrieve similar stuff from DB.

While the implementation can be hashed out and deliberated separately, this PR focuses on the API and some preliminary documentation that goes along with it.


.gitignore
```diff
@@ -135,3 +135,4 @@ dmypy.json
 env3/
 test/aperturedb/db/
 test/input/blobs/
+docs/_static/*.svg
```
luisremis, line 138:
```suggestion
docs/_static/*.svg

```

aperturedb/Collection.py
```diff
@@ -0,0 +1,149 @@
...
+        return Collection
+
+    def show_stats(self) -> None:
+        """
+        **Can be used to show stats about the collection in notebook**
+        """
+        pass
+
+    def display(self,
+                show_bounding_boxes: bool = False,
```
luisremis, line 33: shouldn't this parameters be only present for "Image" collections? they make no sense for a Entity collection for instance.

aperturedb/Collection.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        pass
+
+    def update_properties(self,
+                          related_collection) -> bool:
+        """
+        Propagate changes back to the DB.
+
+        .. note::
+            - The related_collection needs to be of same length as the Collection.
```
luisremis, line 46: what is the data type of a "related_colleciton"? is it also a Collection? or simply a dataframe with property names?

aperturedb/Operations.py
```diff
@@ -0,0 +1,32 @@
+class Operations(object):
+    """
+    **Interface to specify server side processing for images.**
```
luisremis, line 3:
```suggestion
    **Interface to specify server side operations for images.**
```

aperturedb/Repository.py
```diff
@@ -0,0 +1,88 @@
...
+            constraints: The criteria for search, optional
+            operations: Operations before returning the list, optional
+            limit: Maximum number of objects in the collection.
+
+        Returns:
+            Collection of Objects.
+        """
+        return Collection(self._exchange(queries=[self._find_query(constraints)]))
+
+    def ingest_pytorch_dataset(self, dataset: Dataset, converter=None):
```
luisremis, line 69: mmm not sure it make sense to use the Repository abstraction to ingest data this way. I think we should have a separate class (Importer or something like that) with methods for to ingest from different sources.

aperturedb/Repository.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        *Helper to ingest Pytorch Datasets into aperturedb**
+
+        Args:
+            dataset (Dataset): Object of type Dataset from pytorch.datasets
+            converter (_type_, optional): A function that converts a iterable to information for ApertureDB objects. Defaults to None.
+        """
+        pass
+
+    def ingest_kaggle_dataset(self, dataset_ref: str, generator=None, converter=None, kaggle_credentials=None):
```
luisremis, line 79: same here

aperturedb/Repository.py
```diff
@@ -0,0 +1,88 @@
...
+from aperturedb import Connector, Constraints, Operations
+from aperturedb.Collection import Collection
+from torch.utils.data import Dataset
+
+
+class Repository:
+    """**Repository is a store of homogenous entities**
+
+    Notes:
+        - There should be a way to convert to/from pytorch dataset into Data in aperturedb.
```
luisremis, line 11: yes, but probably not in this class

aperturedb/Repository.py
```diff
@@ -0,0 +1,88 @@
+from typing import Dict, List, Optional
+from aperturedb import Connector, Constraints, Operations
+from aperturedb.Collection import Collection
+from torch.utils.data import Dataset
+
+
+class Repository:
+    """**Repository is a store of homogenous entities**
```
luisremis, line 8: not sure what this means. what do you mean here by "entities"?

aperturedb/Collection.py
```diff
@@ -0,0 +1,149 @@
...
+    def retrieve_similar(self, set: str, neigbour_count: int) -> Collection:
+        """
+        Retrieve a collection that has least distance from each of the images in this collection
+
+        Args:
+            set (str): The Descriptor set.
+            neigbour_count (int): The number of neigbours to include in the result.
+
+        Returns:
+            Collection: A new collection of images that comprises of nearest neigbours to this collection.
```
luisremis, line 133:
```suggestion
            Collection: A new collection of objects that comprises of nearest neigbours to this collection.
```

Collection seems like a general abstraction, so these should be similar objects of the type of the original collection.

aperturedb/Collection.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        pass
+
+    def retrieve_by_connection(self, constraint) -> Collection:
```
luisremis, line 137: I think we should make this more general. 
We should have a way to specify specific objects (for instance, entities of the class "Person" with properties "xyz"="something")

aperturedb/Collection.py
```diff
@@ -0,0 +1,149 @@
+from __future__ import annotations
+import pandas as pd
+
+
+class Collection(pd.DataFrame):
```
luisremis, line 5: should this inherit from pd.Dataframe, or simply have a dataframe as a member?

luisremis (commented):
looking good.

I think Collection should be something more general, and we should have different types of collections (each with different parameters for the methods depending on the object type). 
For instance, a collection of images, a collection of bounding boxes, etc. 

What is the plan for that? Do we inherit from Collection? If would be nice to define that as well, and describe how it would all work. If writing an example would help, you can add something on the examples folder.

aperturedb/Repository.py
luisremis, line 8:
```suggestion
    """**Repository is a store of homogenous objects**
```

gsaluja9:
not relavant.

PR #77, opened 2022-04-28
Example 1 : Interaction with pytorch

gsaluja9:
### The main changes are as follows:

1. Per the previous discussion, Datahelper has been refactored into KaggleDataset, which is analogous to pytorch dataset. Usage of loader is decoupled from it.
2. Tests for KaggleDataset mock the kaggle object to replicate the behavior, and do not need actual downloads from kaggle.
3. The 'Examples' have a helper class that is symlinked to each of the example folders. This reduces the boiler plate code to create a connection, or set dynamic arguments from the command line.
4. PytorchDataset is used instead of the lower level batch api.


examples/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+```
+mkdir coco && cd coco && wget http://images.cocodataset.org/zips/val2017.zip && unzip val2017.zip && wget http://images.cocodataset.org/annotations/annotations_trainval2017.zip && unzip annotations_trainval2017.zip
+```
+
+| File | Description | instructions |
+| -----| ------------| -----|
+| AlexNetClassifier.py | Helper code to transorm images before using pretrained alexnet model to classify them | Is not invoked directly |
+| imagenet_classes.txt | The class labels for the outputs from alexnet | used by pytorch_classification.py |
+| prepare_aperturedb.py | Helper to download images from coco datas set, and load them into aperturedb | ``python prepare_aperturedb.py -images_count 100`` |
```
luisremis, line 17:
```suggestion
| prepare_aperturedb.py | Helper to download images from coco dataset, and load them into aperturedb | ``python prepare_aperturedb.py -images_count 100`` |
```

luisremis (commented):
looks great. 

It would be best to put this example in a folder (called "coco-pytorch" or something like that). We will be adding more over time, I already created another PR with other example that is in a separate folder inside the "examples" folder.

luisremis (commented):
looking good. 

What about, instead of example_N, we name the examples with something that will tell the person looking at the examples folder something more meaningful. 
Suggestions:
example_1 -> coco_loader
example_2 -> pytorch_on_coco

I am not a big fan of committing Jupyter notebooks with outputs/state, but I see that will be used to render a doc example, which may be cool. But it is hard to make comments on that file. I am ok with this option for the time being tho.

gsaluja9:
> looking good.
> 
> What about, instead of example_N, we name the examples with something that will tell the person looking at the examples folder something more meaningful. Suggestions: example_1 -> coco_loader example_2 -> pytorch_on_coco
> 
> I am not a big fan of committing Jupyter notebooks with outputs/state, but I see that will be used to render a doc example, which may be cool. But it is hard to make comments on that file. I am ok with this option for the time being tho.

Yea I was not able to quickly figure out a way to make commenting on ipynb happen. Will do some research if that can be done somehow, but I do not have an idea for now.

examples/README.md
```diff
@@ -0,0 +1,28 @@
+# Code examples with aperturedb.
+
+Following are the table of contents for this folder.The instructions to run these scripts are provided alongwith.
+
+A part of Coco validation dataset can be downloaded as follows. This is a prerequisite for running some of the scripts below.
```
luisremis, line 5:
```suggestion
A part of Coco validation dataset can be downloaded as follows. 
This is a prerequisite for running some of the scripts below.
```

examples/image_classification/prepare_aperturedb.py
```diff
@@ -0,0 +1,77 @@
...
+        q[0]["AddImage"]["properties"] = {
+            "dataset_name": "prepare_aperturedb",
+            "id": item[1][0]["image_id"]
+        }
+        return q, [blob]
+
+    # Lets use some images from the coco which are annotated for the purpose of the demo
+    images = []
+    for t in coco_detection:
+        X, y = t
```
luisremis, line 45: what is X? what is y? can we use more descriptive names?

examples/image_classification/prepare_aperturedb.py
```diff
@@ -0,0 +1,77 @@
...
+        stats=True
+    )
+    print(f"Inserted {params.images_count} images to aperturedb")
+
+
+def get_args():
+    parser = argparse.ArgumentParser()
+
+    # Database config
+    parser.add_argument('-db_host', type=str, default="localhost")
```
luisremis, line 64: we should use a dbinfo.py file on the examples, that's we have been doing to avoid replicating this parsing code everywhere. look at the REST api example.

examples/image_classification/pytorch_classification.py
```diff
@@ -0,0 +1,60 @@
+import time
+from aperturedb import Connector
+import AlexNetClassifier as alexnet
+
+db = Connector.Connector(user="admin", password="admin")
```
luisremis, line 5: can we split this file into functions, have a main, and use dbinfo.py to create the connector?

examples/image_classification/pytorch_classification.py
```diff
@@ -0,0 +1,60 @@
...
+        try:
+            start = time.time()
+            res, blobs = db.query(query)
+            print("\rRetrieval performance (imgs/s):",
+                  len(blobs) / (time.time() - start), end="")
+        except Exception as e:
+            print(db.get_last_response_str())
+            raise e
+
+        for id, image in enumerate(blobs):
```
luisremis, line 55: this example is using the batching API, which is more involved. 
The customer request was to use our [PytorchDataset](https://python.docs.aperturedata.io/autoapi/aperturedb/PyTorchDataset/index.html) object.

examples/image_classification/prepare_aperturedb.py
gsaluja9: There is a comment that explains the meaning of the variables, when coco detection (a pytorch object) is instantiated.
It's on line 20-23

examples/image_classification/prepare_aperturedb.py
gsaluja9: but then we'll create multiple copies of dbinfo files (as our examples are in their own subdir).  How should that file be managed ? Maybe putting it in the package ? Lets discuss this ..

examples/README.md
```diff
@@ -0,0 +1,28 @@
+# Code examples with aperturedb.
+
+Following are the table of contents for this folder.The instructions to run these scripts are provided alongwith.
```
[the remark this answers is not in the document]
gsaluja9: We'll need to run all the files through this check, as even README.md in the parent folder has some lines that exceed 80. Do we want to enforce this in code too ?

luisremis (changes requested):
we need to add back the original *Loader.py because we will break data ingestion for our customers. Please put all those files back, and theirs testing. 

Maybe there is a way to keep the interface we had before (with Loaders) and underneath use the *DataCSV? Not sure if worth all the changes and testing. Just adding back *Loader.py should be enough. It will be easier to see the changes as well.

examples/similarity_search/add_faces.py
```diff
@@ -0,0 +1,111 @@
...
+        landmarks_index = pd.read_csv(os.path.join(
+            root, "list_landmarks_align_celeba.csv"))
+        partition_index = pd.read_csv(
+            os.path.join(root, "list_eval_partition.csv"))
+        rows = attr_index.combine_first(bbox_index).combine_first(
+            landmarks_index).combine_first(partition_index)[:helper.params.images_count]
+        print(f"Discovered {len(rows)} items in the original dataset.")
+        return rows
+
+    add_set(helper)
```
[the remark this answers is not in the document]
gsaluja9: ok, will do

gsaluja9:
> we need to add back the original *Loader.py because we will break data ingestion for our customers. Please put all those files back, and theirs testing.
> 
> Maybe there is a way to keep the interface we had before (with Loaders) and underneath use the *DataCSV? Not sure if worth all the changes and testing. Just adding back *Loader.py should be enough. It will be easier to see the changes as well.

Lets discuss this offline once, so that I am sure of the customer requirements.

test/test_kaggle.py
```diff
@@ -0,0 +1,104 @@
...
+        with patch.multiple(KaggleApi,
+                            dataset_list_files = MagicMock(return_value=k),
+                            dataset_download_files = MagicMock(side_effect=dz),
+                            authenticate = MagicMock(
+                                side_effect = self.does_auth)
+                            ) as mocks:
+            dataset = KaggleDataset(
+                dataset_ref=dataset_ref
+            )
+        self.assertNotEqual(len(dataset), 0)
```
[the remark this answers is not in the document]
gsaluja9: BTW, this looks like a comment on old File. we do not have the class anymore.

examples/ExamplesHelper.py
```diff
@@ -0,0 +1,66 @@
...
+        if 'IPKernelApp' not in get_ipython().config:  # pragma: no cover
+            return False
+    except ImportError:
+        return False
+    except AttributeError:
+        return False
+    return True
+
+
+class ExamplesHelper:
```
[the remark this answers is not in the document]
luisremis: sure, it is great to add more functionality (in this case, add the ability to change at run time, which would definately be useful in many cases). Can we just add that functionality to dbinfo.py instead of a new file? We will be reusing dbinfo file all over the place, rather than reusing "ExampleHelper" which is not a good name for something that handles information about the db. 

If you are not happy with the dbinfo name, we can change it for something else, but it would be great to have a unified, explicit way to manage db information that we can reuse for all applications, not only demos. 
I think dbinfo is a good name for it, and it will require the least amount of changes. But happy to discuss if that name is not ok.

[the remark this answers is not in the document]
gsaluja9: The reason I thought that it should be named into something different because in  a few examples, we specify the number of images to be ingested as a command line param. So that was something like a general app configuration, rather then information about db.

I will change the name to dbinfo for now, We can revisit this once there are more examples to go around.

PR #79, opened 2022-04-30
Rest example

luisremis:


examples/rest_api/rest_api.py
```diff
@@ -0,0 +1,134 @@
+import dbinfo
+import requests
+import json
+
+URL = "http://" + dbinfo.DB_HOST + '/api'
+
+
+def dict_query_to_str(dict_query):
+
+    return str(json.dumps(dict_query)).replace("'", '\"')
```
gsaluja9, line 10: the replace is not required, json.dumps converts to JSON with double quotes. Is it for something specefic?

examples/rest_api/rest_api.py
```diff
@@ -0,0 +1,134 @@
...
+
+    return parse_auth(response.text)
+
+
+def query_api(query, st):
+
+    query = dict_query_to_str(query)
+    # print(repr(query))
+
+    response = requests.post(URL, headers = {'Authorization': "Bearer " + st},
```
gsaluja9, line 48: These tokens are to prevent MITM, right? Do we plan to have auth against some auth provider for SSO, AD, etc?

examples/rest_api/rest_api.py
luisremis: we need the double quotes to be scaped, as que query is passed as a string.

examples/rest_api/rest_api.py
luisremis: this is standard https auth. in the future, yes, we would add SSO and other authentication mechanisms.

drewaogle: That's what's nice about the token, it allows a very pluggable auth framework.

gsaluja9 (approved):

PR #80, opened 2022-05-02
remove session renew loop

drewaogle:


luisremis (commented):
Made a few notes, @gsaluja9 will give the approval once tests are passing. Ask him for help on the syntax checker.

aperturedb/Connector.py
```diff
@@ -173,14 +173,14 @@ def _refresh_token(self):
...
-        response, _ = self.query(query)
+        response, _ = self._query(query, [], False)
 
         session_info = response[0]["RefreshToken"]
         if session_info["status"] != 0:
             raise Exception(session_info["info"])
 
-        self.session = Session(session.info["session_token"],
-                               session.info["refresh_token"],
+        self.session = Session(session_info["session_token"],
```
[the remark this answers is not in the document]
drewaogle: yes, my assumption was something changed in the structure somewhere, one of the downsides of python, sadly.

aperturedb/Connector.py
```diff
@@ -234,9 +234,10 @@ def _connect(self):
 
         self.connected = True
 
-    def _query(self, query, blob_array = []):
+    def _query(self, query, blob_array = [], check_status=True):
 
-        self._check_session_status()
```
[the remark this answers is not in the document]
drewaogle: Sure. I noticed `_authenticate` also uses `query` and was thinking that it might be good to have `query` as the public interface and `_query` as the private, where something like `_authenticate` would call `_query` and specify never to refresh - something that might be important if the ttls on sessions were tightened up in the future if we had a problem with too many open connections on the server

test/test_Session.py
```diff
@@ -0,0 +1,40 @@
...
+    def test_sessionRenew(self):
+        '''
+            Verifies that Session renew works
+        '''
+
+        try:
+            db = self.create_connection()
+            # force session token expiry
+            db.session.session_token_ttl = 1
+            print("Connected? {0}".format("yes" if db.connected else "no"))
```
gsaluja9, line 23: can this be an assertion ?

aperturedb/Connector.py
```diff
@@ -280,7 +281,7 @@ def query(self, q, blobs=[]):
             self.response, self.blobs = self._query(q, blobs)
             self.last_query_time = time.time() - start
             return self.response, self.blobs
-        except:
+        except BaseException:
```
gsaluja9, line 284: I was thinking if this could be also propagated back to the user, something like : 
except BaseException as e:
   raise ConnectionError ("ApertureDB disconnected" + e.msg)

gsaluja9 (commented):
Otherwise looks fine, If the cause of actual exception can be propagated up with the connection error, it'd be useful.

test/test_Session.py
drewaogle: Yes, they would work well as asserts

aperturedb/Connector.py
drewaogle: makes sense, I like adding context to traces. this was added by autopep though.

Would you like me to go and add the suggested code you have here?

gsaluja9: Oh I see. We can create it under a new Issue then. I thought you were made the changes.
I ran into a situation with the way we print the message here, without giving back the entire information.

gsaluja9 (approved):
We'll create a better exception handling story to tackle the propagation of error messages.

PR #81, opened 2022-05-03
Example #3 : Similarity search

gsaluja9:


requirements.txt
```diff
@@ -1,5 +1,5 @@
 alabaster==0.7.12
-aperturedb==0.1.1
+aperturedb==0.1.3
```
luisremis, line 2: this should always point to our latest version. can we just indicate that?

gsaluja9: Yea I think we don't care about versions and so many packages here too. I'll clean this up.
Good catch.

examples/similarity_search/add_faces.py
```diff
@@ -0,0 +1,110 @@
...
+import os
+import pandas as pd
+from facenet import generate_embedding, errors
+
+
+search_set_name = "similar_celebreties"
+
+
+def add_set(params):
+    con = Connector.Connector(user=params.db_username,
```
luisremis, line 13: we should use dbinfo.py here.

test/test_kaggle.py
```diff
@@ -0,0 +1,78 @@
...
+import unittest
+from test_Base import TestBase
+import numpy as np
+from PIL import Image
+import io
+import pandas as pd
+import os
+
+
+@unittest.skip("Not intended to be run on CI.")
```
luisremis, line 11: why? it would be great to test the DataHelper class.

aperturedb/DataHelper.py
```diff
@@ -0,0 +1,159 @@
...
+            try:
+                collection = indexer(workdir)
+            except Exception as e:
+                help = f"Failed getting an index to the dataset. Specify custom indexer (https://...) ."
+                raise Exception({
+                    "exception": e,
+                    "help": help
+                })
+
+        con = Connector.Connector(user="admin", password="admin")
```
luisremis, line 113: we cannot have this here. this should be a parameter of the constructor.

aperturedb/DataHelper.py
```diff
@@ -0,0 +1,159 @@
...
+from kaggle.api.kaggle_api_extended import KaggleApi
+from kaggle.models.kaggle_models_extended import File
+
+from aperturedb import Connector
+import zipfile
+
+from aperturedb.ParallelLoader import ParallelLoader
+
+
+class DataHelper(object):
```
luisremis, line 16: I am not sure if I understand what is the purpose of this class. Is it not a general DataHelper, but rather a very specific set of methods to deal with Kaggle. 
We should have a general DataHelper class, and Kaggle being just 1 type of dataset we support.

aperturedb/DataHelper.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    **Class to deal with Data in aperturedb**
```
luisremis, line 18: This is a very loose definition.

aperturedb/DataHelper.py
```diff
@@ -0,0 +1,159 @@
...
+                raise Exception({
+                    "exception": e,
+                    "help": help
+                })
+        else:
+            try:
+                loader.ingest(
+                    generator=list(
+                        map(lambda row: process_row(row), collection.iterrows())),
+                    numthreads=1,
```
luisremis, line 151: why 1 thread and batches of 1? 
we should probably be doing this outside of this class.

gsaluja9:
Merging this into https://github.com/aperture-data/aperturedb-python/pull/77

PR #82, opened 2022-05-16
Add method for counting descriptors on Utils

luisremis:


aperturedb/Utils.py
```diff
@@ -415,3 +415,38 @@ def get_indexed_props(self, class_name, type="entities"):
...
+                "is_connected_to": {
+                    "ref": 1,
+                    "connection_class": "_DescriptorSetToDescriptor"
+                },
+                "results": {
+                    "count": True
+                }
+            }
+        }]
+
```
gsaluja9, line 439:
```suggestion
       total = -1
       try:
            res, _ = self.connector.query(q)
            if not self.connector.last_query_ok():
                self.connector.print_last_response()
            else:
                total = res[1]["FindDescriptor"]["count"]
        except:
            self.connector.print_last_response()
      

        return total

```

aperturedb/Utils.py
```diff
@@ -415,3 +415,38 @@ def get_indexed_props(self, class_name, type="entities"):
...
+                "_ref": 1,
+                "with_name": set_name,
+
+            }
+        }, {
+            "FindDescriptor": {
+                "set": set_name,
+                "is_connected_to": {
+                    "ref": 1,
+                    "connection_class": "_DescriptorSetToDescriptor"
```
gsaluja9, line 432: is "ref": 1 not sufficient ?

gsaluja9 (changes requested):
The code would be easier to read/maintain without multiple return statements.

aperturedb/Utils.py
luisremis: not really, because you can have connection other than _DescriptorSetToDescriptor. We just want to make sure we only count those.

aperturedb/Utils.py
```diff
@@ -415,3 +415,50 @@ def get_indexed_props(self, class_name, type="entities"):
...
+       total = -1
+       try:
+            res, _ = self.connector.query(q)
+            if not self.connector.last_query_ok():
+                self.connector.print_last_response()
+            else:
+                total = res[1]["FindDescriptor"]["count"]
+        except:
+            self.connector.print_last_response()
+
```
luisremis, line 448:
```suggestion
```

gsaluja9 (approved):

PR #85, opened 2022-06-06
Add response handler to parallel queries

luisremis:


aperturedb/ParallelQuery.py
```diff
@@ -30,16 +38,38 @@ def generate_batch(self, data):
 
         return q, blobs
 
+    def call_response_handler(self, r, b):
```
luisremis, line 41: Drew: 
"if hasattr( "response_handler") and callable( self.response_handler ):"

gsaluja9 (approved):

PR #89, opened 2022-06-07
Bugfix : Intermittent failure that manifested in some cases

gsaluja9:
Also simplified dbinfo
TODO : Wether we should restrict the pytest threshold to error. Warning helped discove the issue, and it is conditionally generated.


examples/dbinfo.py
```diff
@@ -1,66 +1,12 @@
...
-            host=self.params.db_host,
-            port=self.params.db_port,
-            user=self.params.db_username,
-            password=self.params.db_password)
-
-    def get_args(self, mandatory_params: Dict):
-        parser = argparse.ArgumentParser()
-
-        # Database config
-        parser.add_argument('-db_host', type=str, default=DB_HOST)
```
drewaogle, line 53: I sometimes run on a different port just to test things - by removing this it seems like there is no way to use prepare_aperturedb.py on a different port now?

gsaluja9: yea, that would be a useful option to specify. But that leads to some extra checks added (Things like checking for ipykernel, etc.) . I wonder if there could be an approach to handle this without adding additional logic into dbinfo.

examples/hello_video/add_videos.py
```diff
@@ -0,0 +1,22 @@
...
+        records_count=helper.params.videos_count,
+    )
+    print(len(dataset))
+
+    loader = ParallelLoader.ParallelLoader(helper.create_connector())
+    loader.ingest(dataset, stats=True)
+
+
+if __name__ == "__main__":
+    main(DBInfo(mandatory_params={
```
luisremis, line 17: I think this is using old DBinfo stuff. Needs change. Maybe add this new example on a separete PR? the rest seems ready for merge.

luisremis (commented):
great. this bug has been there for a long time. thanks!

examples/hello_video/add_videos.py
gsaluja9: this is a file I havent finished working on . Got added by mistake. will remove from the PR

luisremis (approved):
looks good to go!

PR #97, opened 2022-06-17
Add type for object removal

luisremis:
needs testing.


luisremis:
No need, we now have #114 merged.

PR #100, opened 2022-06-27
ConnectionLoader_py_fix

andreim-sv:
### Purpose
Bug/omission fix

### Detail
We use `sys` module without imprting it. Require when an error happens.


luisremis (approved):
great, I think I came across this error at some point. thanks!

PR #103, opened 2022-07-05
Docs and tests for resposne handler.

gsaluja9:
It addresses a bug in reponse handler that called a batch with partial objects with empty values, as it was based on batchsize value.

Added tests for the same.


aperturedb/ParallelQuery.py
```diff
@@ -53,16 +54,13 @@ def do_batch(self, db, data):
...
                     # We could potentially always call this handler function
                     # and let the user deal with the error cases.
 
-                    cmds_per_query = math.ceil(len(r) / self.batchsize)
-                    for i in range(self.batchsize):
-                        start = i * cmds_per_query
-                        end = start + cmds_per_query
-                        self.call_response_handler(
-                            r[start:end], b[start:end])
+                    for i in range(len(r)):
```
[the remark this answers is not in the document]
gsaluja9: oh yes, I will change this to be called once per batch.

test/test_Data.py
```diff
@@ -39,11 +41,12 @@ def insert_data_from_csv(self: TestBase, in_csv_file, rec_count=-1):
...
         print("\n")
 
-    loader = ParallelLoader(self.db)
+    loader = ParallelLoader(self.db,
+                            response_handler=response_handler)
     loader.ingest(data, batchsize=self.batchsize,
                   numthreads=self.numthreads,
                   stats=self.stats)
-    self.assertEqual(loader.error_counter, 0)
+    assert loader.error_counter == 0
```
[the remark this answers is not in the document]
gsaluja9: Ah so this might be a longish explanation.
The way we have used parameterization is based on the pytest. This is actually quite different from the way has such tests. I myself discovered the details while working on this PR. If interested, the docs are here : https://pytest.org/en/7.1.x/how-to/unittest.html

So, to discover pytest type parameterized tests, I could not use a class derived from unittest.TestCase. Hence these tests do not derive from TestBase. This implies that this class does not have self.assert* methods. But the way pytest renders the failures is IMO more useful, and intuitive.

Let me know if you need more information on the same.

aperturedb/ParallelLoader.py
```diff
@@ -10,12 +10,38 @@
...
 class ParallelLoader(ParallelQuery.ParallelQuery):
-    """**Parallel and Batch Loader for ApertureDB**"""
+    """
+    **Parallel and Batch Loader for ApertureDB**
 
-    def __init__(self, db, dry_run=False):
+    It also provides a way for specifying a user defined function to handle the
+    responses of each of the queries executed. This function can be used to process
+    the responses from each of the corresponding queries of a batch in :class:`~aperturedb.Parallelizer.Parallelizer`.
+    It will be called once per batch, and it needs to have 4 parameters:
```
luisremis, line 19:
```suggestion
    It will be called once per query, and it needs to have 4 parameters:
```

aperturedb/ParallelLoader.py
```diff
@@ -10,12 +10,38 @@
...
 
 class ParallelLoader(ParallelQuery.ParallelQuery):
-    """**Parallel and Batch Loader for ApertureDB**"""
+    """
+    **Parallel and Batch Loader for ApertureDB**
 
-    def __init__(self, db, dry_run=False):
+    It also provides a way for specifying a user defined function to handle the
+    responses of each of the queries executed. This function can be used to process
+    the responses from each of the corresponding queries of a batch in :class:`~aperturedb.Parallelizer.Parallelizer`.
```
luisremis, line 18:
```suggestion
    the responses from each of the corresponding queries in :class:`~aperturedb.Parallelizer.Parallelizer`.
```

aperturedb/ParallelLoader.py
```diff
@@ -10,12 +10,38 @@
...
+    Example usage:
 
+    .. code-block:: python
+
+        def process_responses(requests, input_blobs, responses, output_blobs):
+            self.requests.extend(requests)
+            self.responses.extend(responses)
+
+        loader = ParallelLoader(self.db,
+                            response_handler=process_responses)
```
luisremis, line 38:
```suggestion
        loader = ParallelLoader(self.db, response_handler=process_responses)
```

test/conftest.py
```diff
@@ -0,0 +1,57 @@
+import pytest
```
gsaluja9, line 1: This is pytest feature . It automatically imports these fixtures to be used at multiple tests, and can be reused in other tests or fixtures.

Comprehensive docs for the same : https://docs.pytest.org/en/6.2.x/fixture.html

luisremis: cool!

aperturedb/ParallelLoader.py
```diff
@@ -10,12 +10,14 @@
...
-    """**Parallel and Batch Loader for ApertureDB**"""
+    """
+    **Parallel and Batch Loader for ApertureDB**
 
-    def __init__(self, db, dry_run=False):
 
-        super().__init__(db, dry_run=dry_run)
+    """
 
+    def __init__(self, db, dry_run=False, response_handler=None):
```
luisremis, line 19: we no longer need response_handler=None on the constructor, right? Or the idea is to keep it there?

aperturedb/ParallelQuery.py
```diff
@@ -103,7 +159,30 @@ def worker(self, thid, generator, start, end):
...
+        The generator yields a tuple : (array of commands, array of blobs)
+
+        Args:
+            generator (_type_): _description_
+            batchsize (int, optional): _description_. Defaults to 1.
+            numthreads (int, optional): _description_. Defaults to 4.
+            stats (bool, optional): _description_. Defaults to False.
+        """
+        if len(generator) > 0:
+            if isinstance(generator[0], tuple) and isinstance(generator[0][0], list):
```
luisremis, line 173: mmm why is any of this needed? Wouldn't the generators always return [q, [blobs]] ? I will ask you when we chat.

test/test_ResponseHandler.py
```diff
@@ -0,0 +1,159 @@
...
+            "./input/persons.adb.csv", self.requests, self.responses)
+        loader = ParallelLoader(db)
+        loader.ingest(persons, batchsize=99,
+                      numthreads=31,
+                      stats=True)
+        assert loader.error_counter == 0
+
+        dist = persons.df.groupby(persons.df.age // 10 * 10).count()
+        logger.debug(dist["age"])
+
```
luisremis, line 140: great testing, thanks!

luisremis (dismissed):

aperturedb/ParallelLoader.py
luisremis: mmm but why keep it there if we don't plan to use it? Only the response_handler on the generator will be used.

aperturedb/ParallelQuery.py
gsaluja9: The old data loaders do not follow that. Tests will fail.

aperturedb/ParallelLoader.py
gsaluja9: will remove.

aperturedb/ParallelQuery.py
```diff
@@ -54,15 +86,39 @@ def do_batch(self, db, data):
...
+                        end = start + self.commands_per_query
+
+                        blobs_start = i * self.blobs_per_query
+                        blobs_end = blobs_start + self.blobs_per_query
+
+                        b_count = 0
+                        for req, resp in zip(q[start:end], r[start:end]):
+                            for k in req:
+                                # Ref to https://docs.aperturedata.io/parameters/blobs.html
+                                blobs_where_default_true = \
```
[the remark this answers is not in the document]
gsaluja9: Should this be kept for now, or removed right away ?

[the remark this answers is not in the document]
luisremis: we need to keep it until we make the change on athena.

luisremis (approved):
🚀

PR #109, opened 2022-07-20
Add method to retrieve uids for images

luisremis:
Needed for a customer workflow.


luisremis:
Adding high-priority label as this feature is needed for customer workflow.

gsaluja9 (approved):

PR #112, opened 2022-08-07
Fixes error handling when server rejects connection

luisremis:


aperturedb/BBoxLoader.py
```diff
@@ -140,6 +140,7 @@ def generate_batch(self, bbox_data):
             fi = {
                 "FindImage": {
                     "_ref": img_ref,
+                    "blobs": False,
```
luisremis, line 143: this is another unrelated bug fix, where boundingbox loader is unnecessarily retrieving img blobs.

gsaluja9 (approved):
LG

PR #114, opened 2022-08-11
Add method to remove everything from db

luisremis:
This requires ApertureDB support for:

```
{"constraints": {"_uniqueid": ["!=", "0.0.0"]}}
```

### DIFF
--- a/aperturedb/Utils.py
+++ b/aperturedb/Utils.py
@@ -533,3 +533,52 @@ class Utils(object):
             raise e
 
         return total
+
+    def remove_all_objects(self):
+
+        cmd = {"constraints": {"_uniqueid": ["!=", "0.0.0"]}}
+
+        # There is no DeleteDescriptor, but when the sets are removed
+        # all the descriptors are also removed.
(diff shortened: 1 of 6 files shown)


venkat-murty (dismissed):

gsaluja9 (approved):

gsaluja9 (approved):

aperturedb/Utils.py
```diff
@@ -533,3 +533,52 @@ def count_descriptors_in_set(self, set_name):
             raise e
 
         return total
+
+    def remove_all_objects(self):
+
+        cmd = {"constraints": {"_uniqueid": ["!=", "0.0.0"]}}
```
gsaluja9, line 539: btw, what is the implication of not using this constraint ?

luisremis: It is more convenient, because you can use these constraints on the Delete* command directly. Otherwise, we would have to do a Find, declare a ref, and pass a ref to Delete*. Delete must have either a ref or constraints.

PR #128, opened 2022-08-25
Image downloader and uploader to use gcs location instead of local. 

prajapatipreeti:
I am sharing my code that I have updated to add support for downloading and uploading images from gcs location.


luisremis:
Thanks @prajapatipreeti ! We will be including your proposed changes as part of #132

luisremis:
This feature was merged by #132

PR #133, opened 2022-09-08
Minor improvements to Images.py

JoshStoddard:
A few minor changes in support of COCO demo.

The only breaking change is to replace `Images.search_by_id()` with `Images.search_by_property()` (I think this only affects [the COCO polygon demo PR](https://github.com/aperture-data/demos/pull/40))

### DIFF
--- a/aperturedb/Images.py
+++ b/aperturedb/Images.py
@@ -149,7 +149,7 @@ class Images(object):
             find = {
                 "FindImage": {
                     "constraints": {
-                        "_uniqueid": ["==", self.images_ids[idx]]
+                        self.img_id_prop: ["==", self.images_ids[idx]]
                     },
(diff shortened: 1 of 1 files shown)


gsaluja9 (approved):
Looks good. This would cause some notebooks to be changed, right ? I wish there were some tests for Images in general.

JoshStoddard:
> This would cause some notebooks to be changed, right ?

Yep, and only in my pending COCO PR, I'm pretty sure, so nothing should break.

> I wish there were some tests for Images in general.

Yeah, same. It looks like it really only gets exercised in demos. This is another argument for rolling this library into athena python testing: https://aperturedata.slack.com/archives/C02JVQP1WAK/p1662652078539049

PR #144, opened 2022-10-03
Check potential races when updating sessions.

gsaluja9:


aperturedb/Connector.py
```diff
@@ -55,7 +62,8 @@ class Session():
     def valid(self) -> bool:
         session_age = time.time() - self.session_started
 
-        if session_age > self.session_token_ttl:
+        # This triggers refresh if the session is about to expire.
+        if session_age > self.session_token_ttl - 10:
```
drewaogle, line 66: can we make the 10 a variable we define at the top like `SESSION_EXPIRTY_OFFSET_SEC` ?

aperturedb/Connector.py
```diff
@@ -152,39 +160,44 @@ def _authenticate(self, user, password="", token=""):
...
+                                           )
 
     def _check_session_status(self):
-
-        if not self.session:
+        if not self.shared_data.session:
             return
 
-        if not self.session.valid():
+        if not self.shared_data.session.valid():
```
drewaogle, line 174: do we want to think about locking around the test? It could be possible to have a thread start refreshing, then have another enter and do another refresh.

Or perhaps we should retest after locking in `_refresh_token`?

drewaogle (commented):
Looks super useful, just had 2 questions!

aperturedb/Connector.py
gsaluja9: The _refresh_token is only called from within _check_session_status, via _renew_session, which has the logic to try this in case of any failures.
I chose to implement it this way so that we do not have a lock around a bigger section.

drewaogle (approved):
looks great

PR #153, opened 2022-10-17
Release 0.3.3

gsaluja9:
### Change log:

- Preserving the descriptor distance order for images when doing a similarity search
- Working around a Pandas annotation gotcha for Connection Data columns with same names
- Explicitly specifying the unique constraints for the case of ingestion of Bounding boxes and Connections.
- Logging of error, where status is not 0 or -1


luisremis:
this change log is great, what about adding it here?: https://github.com/aperture-data/aperturedb-python/releases/tag/v0.3.3

PR #156, opened 2022-10-19
Change default behaviour of remove_objects on utils

luisremis:


gsaluja9 (approved):

aperturedb/Utils.py
```diff
@@ -582,3 +630,59 @@ def remove_all_objects(self):
             raise e
 
         return True
+
+    def diagnostic(self):
```
luisremis, line 634: this does not work. needs fix. will fix soon. do not merge yet.

PR #171, opened 2022-12-01
Deprecating FindEntity for builtins.

gsaluja9:


test/test_Data.py
```diff
@@ -22,7 +23,8 @@ def remove_entities(self, utils, classes):
 
     def test_Loader(self, utils, insert_data_from_csv):
         classes = ["_Image", "_Descriptor", "Person", "_BoundingBox"]
-        self.remove_entities(utils, classes)
+        self.remove_entities(utils, classes[2:3])
```
luisremis, line 26: remove_all_objects() below should also remove entities of class "Person", right?

test/test_Data.py
```diff
@@ -47,8 +49,7 @@ def test_Loader(self, utils, insert_data_from_csv):
         self.assertEqual(len(data), utils.count_bboxes())
 
     def test_LoaderDescriptorsImages(self, utils, insert_data_from_csv):
-        classes = ["_Image", "_Descriptor", "_DescriptorSet"]
-        self.remove_entities(utils, classes)
+        utils.remove_all_objects()
```
luisremis, line 52:
```suggestion
        self.assertTrue(utils.remove_all_objects())
```
This is valid for all uses of remove_all_objects()

luisremis (approved):
🚀  thanks!

PR #175, opened 2022-12-09
Minor Fixes on Utils

luisremis:
Minor changes and fixes on summary on Utils.py

An important change is this: 
https://github.com/aperture-data/aperturedb-python/pull/175/files#diff-8dc49460bd5a92f5481c3b60675785b408531c14ff242618744a2f6c951602dcR702

Which is causing athena tests to fail because not all the Frame objects are property removed from the db.


luisremis:
@gsaluja9 any idea why this is failing?
Maybe we need the same fix as here?: https://github.com/aperture-data/athena/pull/783/files

luisremis:
@gsaluja9 this is ready for merge I think

gsaluja9 (approved):
nice

PR #176, opened 2022-12-13
Env and Logging Fix

drewaogle:
- Fix depreciation warning for `logging` modules
- Fix `os.getenv` converting to int
- Update python version for testing to 3.10 ( due to https://github.com/actions/runner-images/issues/6399 )
   this causes Ubuntu 22.04 LTS to be used which doesn't have 3.8 by default


gsaluja9 (dismissed):
Cool!

gsaluja9 (approved):

PR #196, opened 2023-01-23
Removing loaders.

gsaluja9:
Removing legacy loaders.
Also ensures that the action run on k8s cluster successfully completed.


luisremis (approved):
🚀

PR #203, opened 2023-02-01
Release 0.4.0

gsaluja9:
### Release notes

- Reconnect on connection drop during a long running connection.
- Tests on python conda env with python 3.7

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -42,11 +42,19 @@ update_version() {
     BUILD_VERSION=$VERSION_BUMP
 }
 
+install_prerequisites() {
+    sudo apt-get update
+    sudo apt-get install -y vim
+}
+
 # Fetch branch
 if [ -z ${BRANCH_NAME+x} ]
 then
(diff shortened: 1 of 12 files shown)


ci.sh
```diff
@@ -42,18 +42,19 @@ update_version() {
     BUILD_VERSION=$VERSION_BUMP
 }
 
+install_prerequisites() {
+    sudo apt-get update
+    sudo apt-get install -y vim
```
luisremis, line 47: why is this needed? this is on the host running ci.

luisremis (approved):
thanks!

ci.sh
gsaluja9: This had me confused also.
https://github.com/aperture-data/aperturedb-python/blob/a913e41278ed0fe3b1181214a0a4e2c11104f3d1/ci.sh#L32

Do you know what command 'ex' is ? It's vim.

PR #219, opened 2023-03-18
Categorize Responses

drewaogle:
Categorizes responses, needed to deal with doing updates where some elements might exist, and a user expects it.

* Adds a list of statuses to consider success to `execute_batch`
*  adds functions to set success criteria for `ParallelQuery` on a class level
*  adds functions to calculate success and number of objects that existed to `ParallelLoader`
*  adds testcase to verify this ( very simple )


aperturedb/ParallelQuery.py
```diff
@@ -69,13 +69,30 @@ def execute_batch(q, blobs, db,
...
-    if isinstance(r, dict) and db.last_response['status'] != 0:
-        logger.error(f"Failed query = {q} with response = {r}")
+
+    statuses = {}
+    if isinstance(r, dict):
+        statuses[r['status']] = [r]
+    elif isinstance(r, list):
+        # add each result to a list of the responses, keyed by the response
+        # code.
+        [statuses.setdefault(result[cmd]['status'], []).append(result)
```
gsaluja9, line 79: This is something that might increase in size for big ingestion runs or if people who run queries using this. To be kept an eye on.

test/test_Success.py
```diff
@@ -0,0 +1,51 @@
...
+
+class TestLoaderSuccess():
+    def assertEqual(self, expected, actual):
+        if expected != actual:
+            raise AssertionError(
+                "Expected {}, got {}".format(expected, actual))
+
+    def assertTrue(self, condition):
+        if not condition:
+            raise AssertionError("Condition not true")
```
gsaluja9, line 17: This would end up in a very obscure message in case of test failing.
pytest assertions have a very nice contextual info added automatically when asserts fail.
https://docs.pytest.org/en/7.1.x/how-to/assert.html

aperturedb/ParallelQuery.py
gsaluja9: Oh you aggregate this from the caller. Should be OK.

test/test_Success.py
drewaogle: should be fixed

PR #225, opened 2023-04-11
Remove the pin on dask.

gsaluja9:
This resolves the CI failures since the release of pandas 2.0.0

### DIFF
--- a/aperturedb/CSVParser.py
+++ b/aperturedb/CSVParser.py
@@ -49,7 +49,8 @@ class CSVParser(Subscriptable):
                 self.filename,
                 blocksize = os.path.getsize(self.filename) // (cores_used * PARTITIONS_PER_CORE))
 
-        if len(self.df) == 0:
+        # len for dask fataframe needs a client.
+        if not use_dask and len(self.df) == 0:
             logger.error("Dataframe empty. Is the CSV file ok?")
 
         self.df = self.df.astype('object')
@@ -57,7 +58,6 @@ class CSVParser(Subscriptable):
         self.validate()
 
     def __len__(self):
-
         return len(self.df.index)
 
     def parse_properties(self, df, idx):
--- a/aperturedb/DaskManager.py
+++ b/aperturedb/DaskManager.py
@@ -17,6 +17,22 @@ logger = logging.getLogger(__name__)
 class DaskManager:
     def __init__(self, num_workers: int = -1):
         self.__num_workers = num_workers
+        # The -1 magic number is to use as many 90% of the cores (1 worker per core).
+        # This can be overridden by the user.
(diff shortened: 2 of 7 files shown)


aperturedb/CSVParser.py
```diff
@@ -49,15 +49,15 @@ def __init__(self, filename, df=None, use_dask=False):
                 self.filename,
                 blocksize = os.path.getsize(self.filename) // (cores_used * PARTITIONS_PER_CORE))
 
-        if len(self.df) == 0:
+        # len for dask fataframe needs a client.
```
drewaogle, line 52:
```suggestion
        # len for dask dataframe needs a client.
```

test/conftest.py
```diff
@@ -81,7 +81,8 @@ def insert_data_from_csv(in_csv_file, rec_count=-1):
                       stats=True,
                       )
         assert loader.error_counter == 0
-        return data
+        # Preserve loader so that dask manager is not auto deleted.
+        return data, loader
```
drewaogle, line 85: I guess this is the reason for the change above? Why do we need to preserve the dask manager?

aperturedb/DaskManager.py
```diff
@@ -17,6 +17,22 @@
...
+        # This can be overridden by the user.
+        # Create a pool of workers.
+        # TODO: see if the same pool can be reused for multiple tasks.
+        workers = self.__num_workers if self.__num_workers != \
+            -1 else int(0.9 * mp.cpu_count())
+
+        self._cluster = LocalCluster(n_workers=workers)
+        self._cluster.shutdown_on_close = False
+        self._client = Client(self._cluster)
+        dask.config.set(scheduler="distributed")
```
drewaogle, line 30: I see you switched the scope of methods for LocalCluster and Client, what was need for the change?

drewaogle (changes requested):

aperturedb/DaskManager.py
gsaluja9: dask distributed , although was doing all the ingestion stuff as the norm, however it was GCing the client and the cluster immediately after the context manager it earlier was confined to. I guess the len (which we use for assertions later) relies on this client, so we keep it in reference till the time were actually done.

test/conftest.py
gsaluja9: It encapsulates the Localcluster, and client.

luisremis (dismissed):
thanks!

test/conftest.py
drewaogle: Ok, sounds good, can you just add something that explains below so someone can know looking at the code later?

great detective work

drewaogle (approved):
Comments look great!

PR #228, opened 2023-04-24
autogenerate indices using parallel loader

### LINKED ISSUE #136: Ingesting CSV data with a `unique_` parameter that is not indexed is slow
It is easy to overlook creating an index for CSV data when ingesting for the first time.  We shoudl at least warn the user when runnign an ingestion script using conditional add on a property that is not indexed that it is going to be slow (particularly when the objects already exist)

JoshStoddard:
When using data loaders, it will attempt to create an index for any property we use for conditional add (ie. with `constraint_` prefix), if it is not already indexed.

Fixes #136

### DIFF
--- a/aperturedb/BBoxDataCSV.py
+++ b/aperturedb/BBoxDataCSV.py
@@ -60,21 +60,21 @@ class BBoxDataCSV(CSVParser.CSVParser):
     def __init__(self, filename, df=None, use_dask=False):
 
         super().__init__(filename, df=df, use_dask=use_dask)
-        if not use_dask:
-            self.props_keys       = [x for x in self.header[5:]
-                                     if not x.startswith(CSVParser.CONSTRAINTS_PREFIX)]
-            self.constraints_keys = [x for x in self.header[5:]
-                                     if x.startswith(CSVParser.CONSTRAINTS_PREFIX)]
 
-            self.img_key = self.header[0]
-            self.command = "AddBoundingBox"
+        self.props_keys = [x for x in self.header[5:]
+                           if not x.startswith(CSVParser.CONSTRAINTS_PREFIX)]
+        self.constraints_keys = [x for x in self.header[5:]
+                                 if x.startswith(CSVParser.CONSTRAINTS_PREFIX)]
+
+        self.img_key = self.header[0]
+        self.command = "AddBoundingBox"
 
     def get_indices(self):
-        return [{
-            "index_type": "entity",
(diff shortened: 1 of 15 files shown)


aperturedb/BBoxDataCSV.py
```diff
@@ -60,14 +60,21 @@ class BBoxDataCSV(CSVParser.CSVParser):
     def __init__(self, filename, df=None, use_dask=False):
 
         super().__init__(filename, df=df, use_dask=use_dask)
-        if not use_dask:
```
gsaluja9, line 63: This check ensures that a CSV file is not read at the first pass when dask mode is used. It's a performance optimization, but it's impact can be observed when using a file with millions of rows. I can explain more offline.

gsaluja9 (changes requested):

aperturedb/BBoxDataCSV.py
JoshStoddard: Thanks, sorry, for got to hit send 🤦 

My interpretation was that this was only making use of the header, which should already be populated by `super().__init__()` and `validation()`. 

In any event, for this PT, I only need the header row.  Is there a way to get that without having to read the whole file?

gsaluja9: I see. I think we can merge it and worry about the optimization part later.

gsaluja9 (approved):

aperturedb/ParallelLoader.py
```diff
@@ -16,6 +16,59 @@ def __init__(self, db, dry_run=False):
...
+
+        if len(new_indices) == 0:
+            return
+
+        logger.info(
+            f"Creating {len(new_indices)} indices: {new_indices}.")
+
+        create_indices = [{"CreateIndex": idx} for idx in new_indices]
+
+        res, _ = self.db.query(
```
luisremis, line 60: same here, we should just use [this method on Utils.py](https://github.com/aperture-data/aperturedb-python/blob/develop/aperturedb/Utils.py#L207).

aperturedb/ParallelLoader.py
```diff
@@ -16,6 +16,59 @@ def __init__(self, db, dry_run=False):
         super().__init__(db, dry_run=dry_run)
         self.type = "element"
 
+    def get_existing_indices(self):
```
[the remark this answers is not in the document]
luisremis: I will create an issue for re-factoring this later, it is not a big deal.

PR #230, opened 2023-05-15
Updating the pin version of protobuf to work with TF

gsaluja9:


drewaogle (approved):
looks good, nice to not have pins

PR #236, opened 2023-06-01
updating docs builder to ubuntu 22

gsaluja9:
This is causing [CI failure](https://github.com/aperture-data/aperturedb-python/actions/runs/5139222602/jobs/9249429931).


luisremis (approved):
👍

PR #241, opened 2023-06-08
Do not run on PRs from main to develop.

gsaluja9:


.github/workflows/pr.yaml
luisremis, line 1: removing this file will cause no CI job to run when a new PR is created. 

develop.yml only runs a job **after** a branch has been merged to `develop`.

luisremis (changes requested):

.github/workflows/pr.yaml
gsaluja9: yea that's right. I will close this. I got confused when looking at this https://github.com/aperture-data/aperturedb-python/actions suggested that there was nothing that ran out of pr. (second column) . but that is just a branch name.

gsaluja9: What do you think about removing push to develop (develop.yml) ? Coming form PR it is already tested. This is just a waste of time

PR #249, opened 2023-06-26
CLI for interacting with aperturedb

gsaluja9:
This provides a few of commonly repeated scripts 
- like generating queries from CSVs.
- Testing connectivity.
- Managing multiple connections for aperturedb instances through cli.

Additionally, not being based on getopt, it has features like bash auto completion, and richer console io.

The most important bit that it provides would be to not have configuration put in as code, use connections across projects (if required).


aperturedb/cli/utilities.py
```diff
@@ -0,0 +1,21 @@
...
+    SUMMARY = "summary"
+    REMOVE_ALL = "remove_all"
+
+
+@app.command()
+def execute(command: CommandTypes):
+    utils = Utils(create_connector())
+    available_commands = {
+        CommandTypes.SUMMARY: utils.summary,
+        CommandTypes.REMOVE_ALL: utils.remove_all_objects
```
luisremis, line 18: We should add wrapper to utils.remove_all_objects call that confirms with the user the deletion. I see myself pressing tab and enter by accident and bye bye data 🥲

aperturedb/cli/transact.py
```diff
@@ -0,0 +1,11 @@
+import json
+
+import typer
+
+app = typer.Typer()
+
+
+@app.command()
+def from_json_file(filepath: str):
```
luisremis, line 9: for the use case of creating indexes, we can just put the query on a .json file and use this. neat. 

I am still unsure what should go on Python scripts and what to move to bash. I think we will learn over time, but if you have an idea, a Readme with your practices to learn from would be good.

aperturedb/Utils.py
```diff
@@ -14,6 +20,51 @@
...
+        Connector: The connector to the database.
+    """
+    all_configs = ls()
+
+    env_config = os.environ.get("APERTUREDB_CONFIG")
+    ac = all_configs["active"]
+    config = all_configs["local"][ac] if "local" in all_configs else all_configs["global"][ac]
+
+    if env_config is not None:
+        # TODO test me.
```
luisremis, line 60: I think this is tested, right?
```suggestion
```

aperturedb/__init__.py
```diff
@@ -40,5 +40,3 @@
 if __version__ != latest_version:
     logger.warning(
         f"The latest version of aperturedb is {latest_version}. You are using version {__version__}. It is recommended to upgrade.")
-    print(
```
luisremis, line 43: thanks!

luisremis (approved):

aperturedb/Utils.py
gsaluja9: This means that the using active config by setting environment was not tested, nor has automated tests.

aperturedb/cli/adb.py
```diff
@@ -0,0 +1,23 @@
...
+app.add_typer(transact.app, name="transact",
+              help="Run a transaction against ApertureDB.")
+
+
+@app.callback()
+def check_context(ctx: typer.Context):
+    if ctx.invoked_subcommand != "config":
+        configure.check_configured(as_global=False) or \
+            configure.check_configured(as_global=True, show_error=True)
+
```
drewaogle, line 20: I looked at the documentation a bit for Typer, and they seem to mainly use this for injecting non-command specific options. They name this `callback` but it seems like  it is a hook that runs around parsing time.

What seems confusing is that they use it to add commands for the main program, which would make it pre-parse but you seem to be using it after that has happened, which would make it post-parse.

Is it both and depends on the args you want it to send?

drewaogle: ok, so I think I've answered my question; the main entry is https://typer.tiangolo.com/tutorial/commands/callback/ but then link tohttps://typer.tiangolo.com/tutorial/commands/context/ which mentions that callback will only be called right before a command by default.

Which is confusing because the first link shows it creating a command line argument --verbose, but I assume the decorator makes it process that.

So now it all makes sense.

aperturedb/cli/ingest.py
```diff
@@ -0,0 +1,60 @@
...
+from aperturedb.VideoDataCSV import VideoDataCSV
+from aperturedb.DescriptorDataCSV import DescriptorDataCSV
+from aperturedb.DescriptorSetDataCSV import DescriptorSetDataCSV
+
+from aperturedb.Utils import create_connector
+from aperturedb.Query import ObjectType
+
+logger = logging.getLogger(__file__)
+app = typer.Typer()
+
```
drewaogle, line 21: it's nice how the typer framework here allows a nice separation of concern so you know that this command requires those parts of the main interface

aperturedb/cli/configure.py
```diff
@@ -0,0 +1,193 @@
...
+from pathlib import Path
+
+import typer
+from typing_extensions import Annotated
+
+from aperturedb.cli.console import console
+from aperturedb.Configuration import Configuration
+
+
+class ObjEncoder(JSONEncoder):
```
drewaogle, line 13: Would be nice to have a explanation of this.

aperturedb/cli/ingest.py
```diff
@@ -0,0 +1,60 @@
...
+        IngestType.IMAGE: ImageDataCSV,
+        IngestType.POLYGON: PolygonDataCSV,
+        IngestType.VIDEO: VideoDataCSV
+    }
+
+    data = ingest_types[ingest_type](filepath, use_dask=use_dask)
+
+    db = create_connector()
+    from aperturedb.cli.console import console
+    console.log(db)
```
drewaogle, line 56: it would be nicer to have the import at top, is there a reason it is here?

test/generateInput.py
```diff
@@ -77,7 +77,7 @@ def generate_blobs_csv():
 def generate_images_csv(multiplier):
 
     multiplier = multiplier // 2
-    path    = "input/images/"
+    path    = "images/"
```
drewaogle, line 80: I liked having a separate "input" directory where the csv was and the the images were below, as were blobs.

gsaluja9: It's still going to end up there. The problem was that the path was relative to the cwd of the running code. With adb, you can run it from arbitrary folder, adjusting the CLI path.

This change mostly sets the image path relative to the csv path, and the DataCSV treats it as such.

aperturedb/cli/ingest.py
gsaluja9: no will adjust that.

aperturedb/cli/configure.py
gsaluja9: I will put the comment.

test/generateInput.py
gsaluja9: BTW, this should be applicable to anything having blobs. I will update the Video and Blobs too.

luisremis (approved):
lgtm! I think we should [addressed this](https://github.com/aperture-data/aperturedb-python/pull/249#discussion_r1267323125) before merging, but it is ok to create an Issue on it and fix it in a subsequent PR if it will be addressed soon. it is scary to be 1 tab away from deleting data.

aperturedb/cli/utilities.py
gsaluja9: this is done.

PR #252, opened 2023-06-27
Fix Img check on ImageCSVData.py

luisremis:


aperturedb/ImageDataCSV.py
```diff
@@ -152,31 +152,62 @@ def load_image(self, filename):
...
+                return True
+            # PNG
+            if img[0] == 137 and img[1] == 80 and img[2] == 78 and img[3] == 71:
+                return True
+            # TIF
+            if img[0] == 73 and img[1] == 73 and img[2] == 42:
+                return True
+
+            # Else, we return false and fail the batch.
             return False
```
gsaluja9, line 189:
```suggestion

```

aperturedb/ImageDataCSV.py
```diff
@@ -152,31 +152,62 @@ def load_image(self, filename):
...
+        except Exception as e:
+            print(e)
```
gsaluja9, line 191:
```suggestion
            logger.exception(e)
```

aperturedb/ImageDataCSV.py
```diff
@@ -152,31 +152,62 @@ def load_image(self, filename):
...
     def load_url(self, url):
         retries = 0
         while True:
             imgdata = requests.get(url)
             if imgdata.ok:
                 imgbuffer = np.frombuffer(imgdata.content, dtype='uint8')
-                if self.check_image and not self.check_image_buffer(imgbuffer):
-                    logger.error("IMAGE ERROR: ", url)
+                if not self.check_image_buffer(imgbuffer):
+                    logger.warning("IMAGE ERROR: ", url)
```
gsaluja9, line 203: This would not show the message on console. Should such an error not be shown to the user ?

gsaluja9 (dismissed):
Minor changes , overall looks good.

aperturedb/ImageDataCSV.py
luisremis: right. Not sure. Usually, I find it more useful to find out about issues after the ingestion, on the error log file generated, rather than polluting the console. At the same time, it is good to know if something is wrong immediately, but on a large ingestion, it is not really useful. 

I will change it back, we can always control the console verbosity with `LOG_CONSOLE_LEVEL=CRITICAL`

aperturedb/ImageDataCSV.py
```diff
@@ -152,31 +152,62 @@ def load_image(self, filename):
...
```
luisremis, line 203:
```suggestion
                    logger.error("IMAGE ERROR: ", url)
```

gsaluja9 (approved):

PR #253, opened 2023-06-29
revert adb version pin

JoshStoddard:
Removes the version pin that was introduced to work around https://github.com/aperture-data/athena/issues/1026


luisremis:
hold this merge until we create a new release of aperturedb. otherwise, it won't work.

luisremis:
We have new release, we should be good to merge this.

gsaluja9 (approved):
looks good.

PR #256, opened 2023-07-18
Add pytest marks and mark checker

drewaogle:
- adds mark_checker which allows gating of imports based on what marks are selected
- adds some marks, and decorations
- uses mark_checker to disable kaggle import unless selected as example.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,4 +1,5 @@
 import pytest
+import mark_checker
 from aperturedb.Connector import Connector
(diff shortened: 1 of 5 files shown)


gsaluja9 (approved):

luisremis:
This is pretty nice, running python tests locally was a bit complicated without this. 

@drewaogle would you add some brief instructions on the description on how markers are specified when running the tests? Thanks! I will give this a try.

PR #259, opened 2023-07-20
Misc bugs in utils.

gsaluja9:


aperturedb/Utils.py
```diff
@@ -617,45 +617,30 @@ def remove_all_objects(self):
 
         cmd = {"constraints": {"_uniqueid": ["!=", "0.0.0"]}}
 
-        queries = [
-            [{"DeleteEntity": cmd}],
-            [{"DeleteImage": cmd}],
-            [{"DeleteVideo": cmd}],
-            [{"DeleteBlob": cmd}],
+        transaction = [
+            {"DeleteEntity": cmd},
```
luisremis, line 621: DeleteEntity should be the last thing. Otherwise, based on the constraints on `cmd`, DeleteEntity will remove all the nodes, including _Image, _Videos, Etc, leaving them file in the file system. 

We need to do DeleteImage, DeleteVideo, etc first, and leave DeleteEntity for last. 

This is a good time to make that fix?

aperturedb/Utils.py
```diff
@@ -617,45 +617,30 @@ def remove_all_objects(self):
...
-            [{"DeleteFrame": cmd}],
+            {"DeleteBoundingBox": cmd},
+            {"DeletePolygon": cmd},
+            {"DeleteFrame": cmd},
+            {"GetSchema": {"refresh": True}}
         ]
 
         try:
-            try:
-                for q in queries:
```
luisremis, line 638: the reason why this was a loop is because it was only practical to remove things on smaller transactions rather than 1 giant transaction. I think that is still the case, but we can try. @venkat-murty any thoughts?

In any case, I think this change is good because this is the end goal: a simple queries that removes everything. 
If we see that it is not practical to remove everything on a single transaction, we can implement something that deletes 10000 things at a time, check the schema, moves on the next 10000s, and iterate.

@bovlb

luisremis (approved):

aperturedb/Utils.py
gsaluja9: OK will do.

test/run_test.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-set -u
-set -e
+#set -u
+#set -e
```
luisremis, line 4: I think this change will make CI pass even after failure on tests.

aperturedb/Utils.py
```diff
@@ -617,45 +617,30 @@ def remove_all_objects(self):
...
             # related objects are deleted.
             # We keep them here until ApertureDB fully implements this.
-            [{"DeleteBoundingBox": cmd}],
-            [{"DeletePolygon": cmd}],
-            [{"DeleteFrame": cmd}],
+            {"DeleteBoundingBox": cmd},
+            {"DeletePolygon": cmd},
+            {"DeleteFrame": cmd},
+            {"DeleteEntity": cmd},
+            {"GetSchema": {"refresh": True}}
```
luisremis, line 634: @venkat-murty assuming all deletes worked, will GetSchema return the new schema after all the deletes?

luisremis (dismissed):

aperturedb/Utils.py
venkat-murty: Yes. If we have any delayed deletes (which we want for descriptors), get schema will return that descriptors exist.

test/run_test.sh
gsaluja9: Thats true. The problem i was trying to address is that script should continue to upload aperturedb stuff to s3 even after failures in prev steps . I suppose I need to do a bash OR or something

Do you have a better idea?

aperturedb/Utils.py
venkat-murty: If we have enough memory then it should not be a problem. If we are deleting n objects then it will create in-memory n objects + an object for every connection + an object for all the connected nodes.

venkat-murty: I would suggest limiting by some using limit = x, safer.

aperturedb/Utils.py
luisremis: ok, so we are good for now. Once we have delayed deletes, we revisit the best method for removing all objects.

aperturedb/Utils.py
venkat-murty: Another problem, a large delete can take a long time (so timeouts are possible) and you have no idea if the system is progressing.

In addition, we (should) would add time limits on queries for system stability in near future.

test/run_test.sh
luisremis: nice, this reminded me I had the same issues. Here is how I solved it: https://github.com/aperture-data/athena/blob/develop/benchmarks/docker/scripts/start.sh#L49 (also, check the comment at the top of the file 🤣 )

aperturedb/Utils.py
luisremis: > I would suggest limiting by some using limit = x, safer.

Yes, loops with a limit would be best, it can be in a followup PR too.

test/run_test.sh
```diff
@@ -27,7 +27,12 @@ RESULT=$?
 if [[ $RESULT != 0 ]]; then
 	echo "Test failed; outputting db log:"
 	if [[ "${APERTUREDB_LOG_PATH}" != "" ]]; then
-		cat -n "${APERTUREDB_LOG_PATH}"/aperturedb.INFO
+		for i in $(ls "${APERTUREDB_LOG_PATH}"/*); do
+			echo "===================="
+			echo "Log file: $i"
+			echo "===================="
+			cat $i
```
drewaogle, line 34: if you use `cat -n` it will print the lines numbers

luisremis (approved):
🚀 thanks!

test/run_test.sh
gsaluja9: done

PR #260, opened 2023-07-21
Dump server logs when tests fail

drewaogle:
example failure: https://github.com/aperture-data/aperturedb-python/actions/runs/5624550195/job/15241597548#step:6:979


luisremis (dismissed):
nice! 🚀

gsaluja9 (approved):
Awesome . just what was needed.
Thanks!

PR #270, opened 2023-08-04
Overridable flag to use blob baths relative to CSVs.

gsaluja9:
Also clean up a bunch of pass through parameters.


luisremis (approved):
🚀

PR #284, opened 2023-08-18
Minimal Python Package

drewaogle:


bovlb:
Did we consider using `extras_require` instead?

drewaogle:
> Did we consider using `extras_require` instead?

yes. The problem is that it would then break every place we currently use the module.

bovlb:
> > Did we consider using `extras_require` instead?
> 
> yes. The problem is that it would then break every place we currently use the module.

Painful I agree, but the result would be cleaner in the long run, especially if we think of other variations (e.g. labelling libraries).  We could deploy it in three steps:
1. Add empty extras for "notebook" or whatever.
2. Switch everywhere that needs to to use the extra.
3. Move non-minimal dependencies into the extras.

drewaogle:
> > > Did we consider using `extras_require` instead?
> > 
> > 
> > yes. The problem is that it would then break every place we currently use the module.
> 
> Painful I agree, but the result would be cleaner in the long run, especially if we think of other variations (e.g. labelling libraries). We could deploy it in three steps:
> 
> 1. Add empty extras for "notebook" or whatever.
> 2. Switch everywhere that needs to to use the extra.
> 3. Move non-minimal dependencies into the extras.

We can always pull that trigger at any point. I think what we can do for now is work on splitting the items into the logical components. A good number of things have a monolithic package and a "pick your own" anyway.

gsaluja9:
As per our discussion the other day, and with our definition of core, complete and dev versions, this [PR](https://github.com/aperture-data/aperturedb-python/pull/312) tries to accomplish these goals in the commonly advocated way.

@drewaogle , I think the overhead of installing pandas and such stuff is minimal, so we should aim to close this. If not, it'd be better to make changes to pyproject.toml.

drewaogle:
> As per our discussion the other day, and with our definition of core, complete and dev versions, this [PR](https://github.com/aperture-data/aperturedb-python/pull/312) tries to accomplish these goals in the commonly advocated way.
> 
> @drewaogle , I think the overhead of installing pandas and such stuff is minimal, so we should aim to close this. If not, it'd be better to make changes to pyproject.toml.

I still think this is useful.

An `aperturedb` package is good for data scientists
An `aperturedb-minimal` package is good for system integrators that want to expose themselves to minimal dependencies.

We don't need to move on this one right now.

luisremis:
I think this is stale and can be closed. please re-open it otherwise.

PR #288, opened 2023-08-21
fix link in python docstring

JoshStoddard:


JoshStoddard:
https://github.com/aperture-data/aperturedb-python/pull/291

PR #293, opened 2023-08-25
Disabling the documentation building workflow.

gsaluja9:
Also fixes some deprecation warnings shown in github actions


.github/workflows/pr.yaml
```diff
@@ -86,14 +86,14 @@ jobs:
...
     - name: Login to DockerHub
-      uses: docker/login-action@v1
+      uses: docker/login-action@v2
       with:
         username: ${{ secrets.DOCKER_USER }}
         password: ${{ secrets.DOCKER_PASS }}
 
     - name: Build documentation
-      run: 'EXCLUDE_TESTING=1 EXCLUDE_DEPLOY=1 NO_PUSH=1 BRANCH_NAME=${{ github.event.pull_request.head.ref }} ./ci.sh'
+      run: 'EXCLUDE_TESTING=1 EXCLUDE_DOCUMENTATION=1 EXCLUDE_DEPLOY=1 NO_PUSH=1 BRANCH_NAME=${{ github.event.pull_request.head.ref }} ./ci.sh'
```
vishakha041, line 98: Does this rule exclude building documentation for a docs build? shouldn't you just remove the rule itself then?

gsaluja9: The idea was to disable it first to see if CI gets affected in any adverse way. It would eventually be removed. This was done when i was doing this on and off. I can get it properly cleaned.

vishakha041 (approved):
Once the CI passes, changes look fine to me. Thanks.

vishakha041:
How will the md files be created if this is disabled? Does docstring have to get executed in the docs workflow now?

PR #296, opened 2023-08-30
fix internal links.

gsaluja9:


aperturedb/Query.py
```diff
@@ -10,7 +10,7 @@
 
 class ObjectType(Enum):
     """
-    Based on : [API description](query_language/Overview/API%20Description)
+    Based on : [API description](/query_language/Overview/API%20Description)
```
vishakha041, line 13:
```suggestion
    Based on : [API description](https://docs.aperturedata.io/query_language/Overview/API%20Description)
```

vishakha041: We don't need to do relative paths there anymore since we will be copying stuff out of this file

vishakha041: Why so? Plus I thot npm was failing due to broken link.

gsaluja9: Yes npm was was broken due to the relative link at the source. That was another level of relative path (in the URI).

The reason we do not want an absolute URL is this documentation ends up at 2 hosts:

1. docs.aperturedata.io (public)
2. silver-bassoon-p1qgm4j.pages.github.io (internal)

If I am browsing the internal docs (let's say proofreading), I should be navigating to the same host.

PR #302, opened 2023-09-07
Add keepalive

bovlb:
This PR adds (client-side) keepalive to `Connector`.  This uses a small third-party library that claims to work on Linux, MacOS, and Windows.  The new feature is enabled by default.

From my testing, queries longer than about five minutes will be disconnected on AWS (and probably on GCP too).  This problem affects `remove_all_objects`, but would also affect large batch imports.  With this new feature enables, I can keep a query running AWS->AWS for an hour without problems.

This feature does increate network traffic slightly, but the only reason I'm aware of to disable it is if you are accessing aperturedb over a cellular connection.  I am therefore enabling it by default.

### DIFF
--- a/aperturedb/Configuration.py
+++ b/aperturedb/Configuration.py
@@ -13,6 +13,7 @@ class Configuration:
     name: str
     use_ssl: bool = True
     use_rest: bool = False
+    use_keepalive: bool = True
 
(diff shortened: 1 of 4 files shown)


gsaluja9 (approved):
lgtm!

PR #306, opened 2023-09-08
update CelebA example

JoshStoddard:
Updates the CelebA similarity search example notebook


gsaluja9:
Thanks for the comments. There were 2 errors in ingest. Any ideas of what those were?

vishakha041:
git is not diffing the notebook.
I like the text changes. Somehow though the performance numbers are way worse and not sure if its ok to put that in the example?

JoshStoddard:
The perf numbers are probably just from running it locally against a debug version of ADB.  The errors complained about NANs in the image data.  Not sure what might have caused it.

vishakha041:
> The perf numbers are probably just from running it locally against a debug version of ADB. The errors complained about NANs in the image data. Not sure what might have caused it.

Could we check in a cleaner version then? @gsaluja9 ?
The text changes are spot on and I like the commenting within code. Just this thing and then gtg.

PR #308, opened 2023-09-20
Addressing ill formed links.

gsaluja9:


aperturedb/Connector.py
```diff
@@ -77,7 +77,7 @@ class Connector(object):
     """
     **Class to facilitate connections with an instance of aperturedb**
 
-    It lets the client execute any query as per the `Native Query specs </category/aperturedb-query-language>`_
+    It lets the client execute any query as per the [Native Query specs](/query_language/Overview/API%20Description)
```
vishakha041, line 80:
```suggestion
    It lets the client execute any JSON query based on the [ApertureDB query language specification](/query_language/Overview/API%20Description)
```

aperturedb/PytorchData.py
```diff
@@ -4,6 +4,21 @@
 
 
 class PytorchData(Subscriptable):
```
vishakha041, line 6:
```suggestion
class PyTorchData(Subscriptable):
```

aperturedb/PytorchData.py
```diff
@@ -4,6 +4,21 @@
...
+    """
+    **Class to wrap around a Dataset retrieved from torchvision.datasets**
```
vishakha041, line 8: Do we have a link to refer here for torch vision?

aperturedb/PytorchData.py
```diff
@@ -4,6 +4,21 @@
...
+
+    The dataset in this case can be iterated over to yield atomic records.
```
vishakha041, line 10: What does atomic records here mean?

vishakha041 (changes requested):
A few changes and this looks good

JoshStoddard (commented):
👍 LGTM

aperturedb/PytorchData.py
gsaluja9: will add

aperturedb/PytorchData.py
gsaluja9:
```suggestion
    The dataset in this case can be iterated over.
```

PR #310, opened 2023-09-26
Add remove_all_indexes

bovlb:
This method will be useful in improving the performance of coco ingestion.


aperturedb/Utils.py
```diff
@@ -663,6 +663,42 @@ def count_descriptors_in_set(self, set_name):
...
+                return False
+        except BaseException as e:
+            logger.exception(e)
+            return False
+
+        return True
+
     def remove_all_objects(self):
 
         cmd = {"constraints": {"_uniqueid": ["!=", "0.0.0"]}}
```
luisremis, line 704: Should we make `remove_all_objects` call `remove_all_indexes` before removing the objects?

gsaluja9 (approved):
looks good.

aperturedb/Utils.py
bovlb: Maybe.  I'm cautious about changing existing behaviour in case someone is relying on it.  I did think about adding a `remove_all_indexes` flag to `remove_all_objects`.

bovlb: I'm going to mark this resolved because we can make this enhancement in a future PR, and I want to go ahead and experiment with deleting indexes.

aperturedb/Utils.py
```diff
@@ -663,6 +663,42 @@ def count_descriptors_in_set(self, set_name):
...
 
+    def remove_all_indexes(self):
+        """Remove all indexes from the database.
+
+        This may improve the performance of remove_all_objects.
+        It may improve or degrade the performance of other operations.
+
+        Note that this only removes populated indexes.
+        """
+        def find_indexes(schema):
```
gsaluja9, line 674: Sorry this was an oversight. There's already a similar implementation 
https://github.com/aperture-data/aperturedb-python/blob/70fb10399128a6c525aa2a9990cf1f689861eaf4/aperturedb/ParallelLoader.py#L22

and remove index itself exists in utils:
https://github.com/aperture-data/aperturedb-python/blob/70fb10399128a6c525aa2a9990cf1f689861eaf4/aperturedb/Utils.py#L240

Can this all be reconciled? @bovlb

gsaluja9: Utils might be a good place to put this in.

bovlb: I'll consolidate `get_existing_indices` with `find_indexes`.
I didn't want to use `_remove_index` because it generates a query per index.

PR #311, opened 2023-09-27
Fixing remove_all invocation.

gsaluja9:
The utils commands with a pipelined confirmation were not executing because of a bug.
Also adressed a sporadic failure while ingesting a small file with dask.


aperturedb/cli/utilities.py
```diff
@@ -29,9 +30,9 @@ def execute(command: CommandTypes):
     available_commands = {
         CommandTypes.SUMMARY: utils.summary,
         CommandTypes.REMOVE_ALL: lambda: confirm(
-            CommandTypes.REMOVE_ALL) and utils.remove_all_objects,
+            CommandTypes.REMOVE_ALL) and utils.remove_all_objects()
```
gsaluja9, line 33:
```suggestion
            CommandTypes.REMOVE_ALL) and utils.remove_all_objects(),
```

test/conftest.py
```diff
@@ -77,7 +77,7 @@ def insert_data_from_csv(db, request):
     used to parse semantics of the .csv file
     """
     def insert_data_from_csv(in_csv_file, rec_count=-1, expected_error_count=0, loader_result_lambda=None):
-        if rec_count > 0 and rec_count < 80:
+        if rec_count > 0 and rec_count < 80 or "descriptorset" in in_csv_file:
             request.param = False
             print("Not enough records to test parallel loader. Using serial loader.")
```
drewaogle, line 82: seems like having descriptorset will also cause the error message as well, should add that to the print?

gsaluja9: The reason is still the same. The CSV we have is tiny one. 
Even then This print in tests is hardly of use, and if we still try to use tiny files in dataCSV, they get [caught](https://github.com/aperture-data/aperturedb-python/blob/233f9397a64d651c6011a2df3043f2ef881b5db2/aperturedb/CSVParser.py#L70).

gsaluja9: I'm going to mark this resolved for now.

bovlb (approved):
First part looks good, but I don't understand the `descriptorset` part.

drewaogle:
> First part looks good, but I don't understand the `descriptorset` part.

iirc, we can't split up descriptorset loads because the way they are created

PR #312, opened 2023-09-27
Feature/packaging

gsaluja9:


pyproject.toml
```diff
@@ -0,0 +1,81 @@
...
+"Bug Reports" = "https://github.com/aperture-data/aperturedb-python/issues"
+
+[project.optional-dependencies]
+# This is used when we build the docker image for notebook
+notebook = [
+    "torch",
+    "torchvision",
+    "tensorflow",
+    "facenet-pytorch",
+]
```
drewaogle, line 46: looks good. when we make this split what demos will we have to upgrade, and will we need to inform some customers?

drewaogle: ok, I think everything is good here now?

drewaogle (approved):
lgtm

Issue #258, opened 2023-07-18
TCP Connector not recovering from a failed connection


luisremis:
It seems the re-connection mechanism on Connector.py is not working properly and recovering from an error. 

From the `watchdog` deployment on `web-develop` cluster logs:

```
watchdog-aperturedata-6799d7dd5c-h2m56 Checking: coco.datasets.cloud.aperturedata.io - TCP...too many values to unpack (expected 2)                                              │
│ watchdog-aperturedata-6799d7dd5c-h2m56 Already notified: coco.datasets.cloud.aperturedata.io, False                                                                              │
│ watchdog-aperturedata-6799d7dd5c-h2m56 Checking: coco.datasets.cloud.aperturedata.io - REST... 4.1446685791015625 ms                                                             │
│ watchdog-aperturedata-6799d7dd5c-h2m56 2023-07-18 21:20:24,323 : CRITICAL : aperturedb.Connector : 140449908715968 : 351 : too many values to unpack (expected 2)                │
│ watchdog-aperturedata-6799d7dd5c-h2m56 Checking: yfcc100m.datasets.cloud.aperturedata.io - TCP...ApertureDB disconnected                                                         │
│ watchdog-aperturedata-6799d7dd5c-h2m56 Already notified: yfcc100m.datasets.cloud.aperturedata.io, False                                                                          │
│ watchdog-aperturedata-6799d7dd5c-h2m56 Checking: yfcc100m.datasets.cloud.aperturedata.io - REST... 4.317760467529297 ms                                                          │
│ watchdog-aperturedata-6799d7dd5c-h2m56 Checking: cifar100.datasets.cloud.aperturedata.io - TCP... 1.5075206756591797 ms                                                          │
│ watchdog-aperturedata-6799d7dd5c-h2m56 Checking: cifar100.datasets.cloud.aperturedata.io - REST... 4.980802536010742 ms                                                          │
│ watchdog-aperturedata-6799d7dd5c-h2m56 Checking: badgertech.cloud.aperturedata.io - TCP... 0.7762908935546875 ms                                                                 │
│ watchdog-aperturedata-6799d7dd5c-h2m56 Checking: badgertech.cloud.aperturedata.io - REST... 4.493951797485352 ms                                                                 │
│ watchdog-aperturedata-6799d7dd5c-h2m56 Sleeping...                                                                                                                               │
│ watchdog-aperturedata-6799d7dd5c-h2m56 Woke up.                                                                                                                                  │
[...]

PR #315, opened 2023-09-29
Re establish connection on errors.

gsaluja9:
This would handle scenarios where a long running connection would be checked for errors, with logic to recover.
Reported as in #258

### DIFF
--- a/aperturedb/Configuration.py
+++ b/aperturedb/Configuration.py
@@ -14,10 +14,10 @@ class Configuration:
     use_ssl: bool = True
     use_rest: bool = False
     use_keepalive: bool = True
-    retry_connect_interval_seconds: int = 10
+    retry_connect_interval_seconds: int = 1
     # Max number of attempts to retry the initial connection (0 means infinite)
     # This is useful when the aperturedb server is not ready yet.
-    retry_connect_max_attempts: int = 0
+    retry_connect_max_attempts: int = 3
 
     def __repr__(self) -> str:
         mode = "REST" if self.use_rest else "TCP"
--- a/aperturedb/Configuration.py
+++ b/aperturedb/Configuration.py
@@ -14,6 +14,10 @@ class Configuration:
     use_ssl: bool = True
     use_rest: bool = False
     use_keepalive: bool = True
+    retry_connect_interval_seconds: int = 10
+    # Max number of attempts to retry the initial connection (0 means infinite)
+    # This is useful when the aperturedb server is not ready yet.
+    retry_connect_max_attempts: int = 0
 
     def __repr__(self) -> str:
         mode = "REST" if self.use_rest else "TCP"
--- a/aperturedb/Connector.py
+++ b/aperturedb/Connector.py
@@ -358,11 +358,15 @@ class Connector(object):
                         response_blob_array = [b for b in querRes.blobs]
                         self.last_response = json.loads(querRes.json)
                         break
-            except ssl.SSLError as e:
+            except ssl.SSLError as ssle:
                 # This can happen in a scenario where multiple
                 # processes might be accessing a single connection.
                 # The copy does not make usable connections.
-                logger.warning(f"Socket error on process {os.getpid()}")
+                logger.exception(ssle)
+                logger.warning(f"SSL error on process {os.getpid()}")
+            except OSError as ose:
+                logger.exception(ose)
+                logger.warning(f"OS error on process {os.getpid()}")
             tries += 1
             logger.warning(
                 f"Connection broken. Reconnectng attempt [{tries}/3] .. PID = {os.getpid()}")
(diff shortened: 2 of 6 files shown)


gsaluja9:
This is blocked on this : https://github.com/aperture-data/athena/pull/1205

aperturedb/Configuration.py
```diff
@@ -14,6 +14,10 @@ class Configuration:
     use_ssl: bool = True
     use_rest: bool = False
     use_keepalive: bool = True
+    retry_connect_interval_seconds: int = 10
+    # Max number of attempts to retry the initial connection (0 means infinite)
+    # This is useful when the aperturedb server is not ready yet.
+    retry_connect_max_attempts: int = 0
```
drewaogle, line 20: lets not do infinite as the default. I think expect a library to return an error at some point.

test/test_Session.py
```diff
@@ -79,3 +82,139 @@ def mock_send_msg(msg):
...
+            new_db = Connector(
+                dbinfo.DB_TCP_HOST,
+                dbinfo.DB_TCP_PORT,
+                dbinfo.DB_USER,
+                dbinfo.DB_PASSWORD,
+                retry_connect_max_attempts=3,
+                retry_connect_interval_seconds=0)
+        except Exception as e:
+            # Check the exception is not an obscure one.
+            assert str(e).startswith("Could not connect to apertureDB server:")
```
drewaogle, line 107: nice test!

drewaogle (commented):
just the question about the default otherwise lgtm

aperturedb/Configuration.py
gsaluja9: I thought so too myself, but was convinced otherwise thinking of attaining analogous behavior and a current customer (watchdog).

1. So ConnectorRest does not encapsulate a connection. Not doing like this would introduce 2 separate default behaviors using rest or not.
2. The watchdog .... with increasing number of URLs, it'll be hard to ensure that we start it after all servers are up, and this is the case that got @luisremis because they anyways did not care (or write code to handle the response after creating the connector. I think this behavior can be explicitly documented but it gives the connectors a parity on how they init.

If you are convinced then, I can update the docs to highlight this, otherwise lets discuss this more.

aperturedb/Connector.py
```diff
@@ -383,8 +425,9 @@ def query(self, q, blobs=[]):
             self.last_query_time = time.time() - start
             return self.response, self.blobs
         except BaseException as e:
-            logger.critical(e)
-            raise ConnectionError("ApertureDB disconnected")
+            logger.critical(traceback.format_exc(limit=10))
```
bovlb, line 428: Why not use `exc_info` here?

aperturedb/Connector.py
```diff
@@ -383,8 +425,9 @@ def query(self, q, blobs=[]):
...
+            logger.critical("\r\n".join(traceback.format_stack(limit=10)))
```
bovlb, line 429: Why not use `stack_info` here?

test/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3'
 
 services:
   aperturedb:
-    image: aperturedata/aperturedb
+    image: aperturedata/aperturedb:vdevelop
```
bovlb, line 5: Why the change in image?

aperturedb/Configuration.py
gsaluja9: OK, I checked ConnectorRest and I stand corrected. Even that tries to connect in init, and bails out if server is not up.
I shall make this from infinite to 3, and 1 second for retry interval as default. 
Thanks for pointing this out.

test/test_Session.py
gsaluja9: Discussed this in IM.

test/docker-compose.yml
gsaluja9: Same as above

drewaogle: cool.

aperturedb/Configuration.py
drewaogle: good to know!

We should put this in docs so we know our theory of operation on this stuff.

aperturedb/Connector.py
gsaluja9: will address.

test/docker-compose.yml
luisremis: [v0.15.16](https://github.com/aperture-data/athena/compare/v0.15.15...v0.15.16) is ready to go!

luisremis (dismissed):
🚀

test/docker-compose.yml
drewaogle: I think we can resolve this

PR #319, opened 2023-10-17
Apply on the fly changes while ingestion.

gsaluja9:
This introduces a concept of pipeline to augment a dataset while being ingested.

An example use case that was mixing a lot of concerns while ingestion was [CelebADataKaggle](https://github.com/aperture-data/aperturedb-python/blob/bfaee4b9cbc410be844fe78aad5ff37102c2c81d/examples/CelebADataKaggle.py).

now all the stuff that needs to go in is built as a [pipeline](https://github.com/aperture-data/aperturedb-python/blob/bfaee4b9cbc410be844fe78aad5ff37102c2c81d/examples/similarity_search/add_faces.py).This was a multiple user ask, makes the concepts more understandable in docs

TODO : 
Add more docs explaining this.
Add tests.


aperturedb/cli/ingest.py
```diff
@@ -25,6 +30,70 @@
...
+    """
+    db = create_connector()
+    loader = ParallelLoader(db)
+
+    # not sure if this is the best way to do this.
+    # Hence the BETA tag.
+    module_name = os.path.basename(filepath)[:-3]
+    spec = importlib.util.spec_from_file_location(module_name, filepath)
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[module_name] = module
```
drewaogle, line 76: I think importlib is the defacto way to load modules dynamically. Why not explain explicitly what you're doing here

aperturedb/transformers/facenet.py
```diff
@@ -0,0 +1,25 @@
+from facenet_pytorch import MTCNN, InceptionResnetV1
+from PIL import Image
+import torch
```
drewaogle, line 3: if we aren't going to require these modules in requirements we might want to wrap the import with a try

aperturedb/transformers/image_properties.py
```diff
@@ -0,0 +1,29 @@
...
+        super().__init__(data)
+
+    def getitem(self, subscript):
+        x = self.data[subscript]
+        try:
+            # x is a transaction that has an add_image command and a blob
+            for ic in self._add_image_index:
+                # Compute the dynamic properties and apply them to metadata
+                x[0][ic]["AddImage"]["properties"]["adb_image_size"] = len(
+                    x[1][ic])
```
drewaogle, line 19: very cool, was thinking this would be nice to have. these could be useful in scripts as well as the commandline

kind of like a "middleware" .. 

```
qr = ParallelQuery
qr.ingest( csv, transformers = [ ImageProps, Resizer ] )
```

aperturedb/transformers/facenet.py
gsaluja9: fair point. will do so.

aperturedb/transformers/image_properties.py
gsaluja9: This reads so easy on the eyes.
I am not yet doing it, as it needs to be tested on DaskManager.

aperturedb/transformers/image_properties.py
```diff
@@ -0,0 +1,32 @@
...
+
+                # Compute the image dimensions.
+                pil_image = Image.open(io.BytesIO(x[1][ic]))
+                x[0][ic]["AddImage"]["properties"]["adb_image_width"] = pil_image.width
```
vishakha041, line 26: I like using these variables BUT I remember seeing COCO images get added with width, height which causes a duplicate. Do you know how to address that?

aperturedb/transformers/transformer.py
```diff
@@ -0,0 +1,76 @@
...
+        .
+        .
+
+        dataset = CelebADataKaggle()
+
+        # Here's a pipeline that adds extra properties to the celebA dataset
+        dataset = CommonProperties(
+            dataset,
+            adb_data_source="celebA",
+            adb_main_object="Face",
```
vishakha041, line 37: Awesome.

aperturedb/transformers/transformer.py
```diff
@@ -0,0 +1,76 @@
...
+        .
+        .
+        .
+
+        dataset = CelebADataKaggle()
+
+        # Here's a pipeline that adds extra properties to the celebA dataset
+        dataset = CommonProperties(
+            dataset,
+            adb_data_source="celebA",
```
vishakha041, line 36:
```suggestion
            adb_data_source="kaggle-celebA",
```

aperturedb/transformers/transformer.py
```diff
@@ -0,0 +1,76 @@
...
+        x = self.data[0]
+        self._queries = len(x[0])
+        self._blobs = len(x[1])
+        self._blob_index = []
+        self._add_image_index = []
+
+        bc = 0
+        for i, c in enumerate(x[0]):
+            command = list(c.keys())[0]
+            if command in ["AddImage", "AddDescriptor", "AddVideo", "AddBlob"]:
```
vishakha041, line 63: Should we create an issue to introduce this feature for videos / blobs / other objects too after this PR is merged?

examples/similarity_search/add_faces.py
```diff
@@ -1,23 +1,35 @@
...
-        embedding_generator=generate_embedding,
-        search_set_name=search_set_name
-    )
+
+    dataset = CelebADataKaggle()
+
+    # Here's a pipeline that adds extra properties to the celebA dataset
+    dataset = CommonProperties(
+        dataset,
+        adb_data_source="celebA",
```
vishakha041, line 21:
```suggestion
        adb_data_source="kaggle-celebA",
```

vishakha041: Just want to be more clear where data came from but ok if you want to leave it at celebA

examples/similarity_search/add_faces.py
```diff
@@ -1,23 +1,35 @@
...
+        adb_main_object="Face",
+        adb_annoted=True)
+
+    # some useful properties for the images
+    dataset = ImageProperties(dataset)
+
+    # Add the embeddings generated through facenet.
+    dataset = FacenetPytorchEmbeddings(dataset)
```
vishakha041, line 29: Is this example now duplicated here and in our Python package? Which one does docs use?

aperturedb/TensorflowData.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Example subclass: [Cifar100DataTensorflow](https://github.com/aperture-data/aperturedb-python/blob/develop/examples/Cifar100DataTensorflow.py)
+
+    """
+
+    def __init__(self, dataset: tf.data.Dataset) -> None:
+        self.loaded_dataset = [t for t in dataset]
+
+    def getitem(self, idx: int):
+        return self.generate_query(idx)
```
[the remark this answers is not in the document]
gsaluja9: Yes , that's correct. this follows on the lines of KaggleData and PytorchData. Will split off.

aperturedb/transformers/image_properties.py
drewaogle: yeah, for sure later. I'll add a issue for feature

https://github.com/aperture-data/aperturedb-python/issues/326

aperturedb/transformers/common_properties.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+
+    def __init__(self, data: Subscriptable, **kwargs) -> None:
+        """
+        Args:
+            data: Subscriptable object
+            adb_data_source: Data source for the data
+            adb_timestamp: Timestamp for the data
+            adb_main_object: Main object for the data
+            adb_annoted_by: Annoted by for the data
```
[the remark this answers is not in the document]
gsaluja9: I am going to drop this out for now in the iteration.
We can decide how to make this happen later. I do not have any concrete ideas for now.

aperturedb/transformers/transformer.py
gsaluja9: yes. I have not evaluated other blob types as yet.

examples/similarity_search/add_faces.py
gsaluja9: The difference in this and our adb ingest is the FacenetEmbeddings. docs will not show FacenetEmbeddings as an option, this although they will callout that such a class exists if anyone wants to 

- use it, as it is.
- refer to it to build something custom.

aperturedb/transformers/image_properties.py
gsaluja9: I am not sure if this can be consolidated because of the down stream consumers of these properties.
Leaving them in right now. We should discuss this more.

aperturedb/transformers/common_properties.py
```diff
@@ -0,0 +1,42 @@
...
+class CommonProperties(Transformer):
+    """
+    This applies some common properties to the data.
+    """
+
+    def __init__(self, data: Subscriptable, **kwargs) -> None:
+        """
+        Args:
+            data: Subscriptable object
+            adb_data_source: Data source for the data
```
vishakha041, line 15: Another thing I ran into just now. How about indexing for some of these, especially data_source? How do we keep it optional?

gsaluja9: Check if index exists, and create if not ? Would that be OK. It's in line with DataCSVs which create indexes for properties with constraint properties.

vishakha041: That sounds good. ADd it here ?

aperturedb/transformers/image_properties.py
vishakha041: I created an issue for this

aperturedb/transformers/transformer.py
```diff
@@ -0,0 +1,76 @@
...
+        .
+
+        dataset = CelebADataKaggle()
+
+        # Here's a pipeline that adds extra properties to the celebA dataset
+        dataset = CommonProperties(
+            dataset,
+            adb_data_source="kaggle-celebA",
+            adb_main_object="Face",
+            adb_annoted=True)
```
vishakha041, line 38:
```suggestion
            adb_annotated=False)
```

vishakha041: we dont have bounding boxes in the celebA dataset, do we?

aperturedb/transformers/transformer.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # some useful properties for the images
+        dataset = ImageProperties(dataset)
+
+        # Add the embeddings generated through facenet.
```
vishakha041, line 43: Won't this create embeddings even if you just want to add simple properties but not embeddings?

aperturedb/transformers/common_properties.py
```diff
@@ -0,0 +1,42 @@
...
+        self.adb_timestamp = kwargs.get("adb_timestamp", None)
+        self.adb_main_object = kwargs.get("adb_main_object", None)
+
+    def getitem(self, subscript):
+        x = self.data[subscript]
+        try:
+            # x is a transaction that has an add_image command and a blob
+            for ic in self._add_image_index:
+                # Set the static properties, if explicitly set
+                if self.adb_data_source:
```
vishakha041, line 32: Also, how about standardized adb_image_id please?

examples/Foo.py
```diff
@@ -0,0 +1,16 @@
+from aperturedb.transformers.transformer import Transformer
+
+
+class Foo(Transformer):
```
vishakha041, line 4: Seriously! Foo bar! we compiler ppl now? :)

examples/similarity_search/add_faces.py
```diff
@@ -1,23 +1,35 @@
...
-    )
+
+    dataset = CelebADataKaggle()
+
+    # Here's a pipeline that adds extra properties to the celebA dataset
+    dataset = CommonProperties(
+        dataset,
+        adb_data_source="kaggle-celebA",
+        adb_main_object="Face",
+        adb_annoted=True)
```
vishakha041, line 23:
```suggestion
        adb_annotated=False)
```

vishakha041: I would put true in coco

examples/similarity_search/add_faces.py
```diff
@@ -1,23 +1,35 @@
...
+
+    # some useful properties for the images
+    dataset = ImageProperties(dataset)
+
+    # Add the embeddings generated through facenet.
+    dataset = FacenetPyTorchEmbeddings(dataset)
```
vishakha041, line 29: Oh this is repeated because we have this in examples and somewhere else but I am not clear why.
Also, same questuon on creating embeddings if one wants only image props

vishakha041 (changes requested):
Minor changes and questions. Looks almost ready to merge

examples/similarity_search/add_faces.py
gsaluja9: So, this is true, because CelebA has bounding boxes and key points.

examples/Foo.py
gsaluja9: This is just to give an example in docs as to what this is.

aperturedb/transformers/common_properties.py
gsaluja9: done.

examples/Foo.py
gsaluja9: I've updated some docstrings. will add more description once this is merged.

examples/similarity_search/add_faces.py
gsaluja9: anyways, removed this as we are holding back on annotated*

examples/similarity_search/add_faces.py
gsaluja9: yes repeated in how examples have it, and now the CLI can pick and choose these to build a pipeline.
Any combination is allowed through CI.

aperturedb/cli/ingest.py
```diff
@@ -13,33 +16,151 @@
...
+from aperturedb.Utils import import_module_by_path
+
 logger = logging.getLogger(__file__)
 app = typer.Typer()
 
 
 IngestType = Enum('IngestType', {k: str(k) for k in ObjectType._member_names_})
 
 
+class TransformerType(str, Enum):
```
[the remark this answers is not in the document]
gsaluja9: added.

aperturedb/cli/ingest.py
```diff
@@ -13,33 +16,151 @@
...
+        help="Number of workers for ingestion")] = 1,
+    stats: Annotated[bool, typer.Option(
+        help="Show realtime statistics with summary")] = True,
+    use_dask: Annotated[bool, typer.Option(
+        help="Use dask based parallelization")] = False,
+    ingest_type: Annotated[IngestType, typer.Option(
+        help="Parser for CSV file to be used")] = IngestType.IMAGE,
+    blobs_relative_to_csv: Annotated[bool, typer.Option(
+        help="If true, the blob path is relative to the csv file")] = True,
+    transformer: Annotated[Optional[List[TransformerType]], typer.Option(
```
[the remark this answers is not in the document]
gsaluja9: yes. the user may mix and match now.

aperturedb/transformers/common_properties.py
gsaluja9: done

vishakha041 (dismissed):
Gautam addressed review comments live.

PR #320, opened 2023-10-18
Internal ref ids start at 10000

bovlb:
The aperturedb API permits refs to be in the range [1,100000), and uses higher numbers for internal purposes.  Because of a recent change in the aperturedb API, we are now enforcing this restriction, whereas previously violations were quietly ignored.  The CSV loader wants to avoid ref collision when batching, so assigns refs serially, modulo a limit.  Because these refs are sometimes assigned in pairs, we were going over this limit.  This change ensures we stay under the limit.

Note that there is still an outstanding problem because the CSV loader is merely hoping that this modulo trick is sufficient to ensure that batches are semantically coherent.  A better approach might be to have the batching process itself remap ref numbers on the fly.

### DIFF
--- a/aperturedb/BBoxDataCSV.py
+++ b/aperturedb/BBoxDataCSV.py
@@ -83,7 +83,7 @@ class BBoxDataCSV(CSVParser.CSVParser):
 
         img_id = self.df.loc[idx, self.img_key]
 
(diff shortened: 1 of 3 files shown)


gsaluja9 (approved):
Thanks. I ran into this very recently :)

PR #330, opened 2023-10-25
Generator to move data from TF datasets into aperturedb

gsaluja9:
This follows the line of classes like KaggleData and Pytorch Data in apertureDB

### DIFF
--- a/aperturedb/TensorFlowData.py
+++ b/aperturedb/TensorFlowData.py
@@ -31,7 +31,7 @@ class TensorFlowData(Subscriptable):
 
     def generate_query(self, idx: int) -> Tuple[List[dict], List[bytes]]:
         """
-        **Takes information from one atomic record from the Data and converts it to Query for apertureDB**
+        **Takes information from one atomic record from the Data and converts it to Query for ApertureDB**
 
         Args:
             idx (int): index of the record in collection.
--- a/aperturedb/TensorFlowData.py
+++ b/aperturedb/TensorFlowData.py
@@ -21,13 +21,13 @@ class TensorFlowData(Subscriptable):
     """
 
     def __init__(self, dataset: tf.data.Dataset) -> None:
-        self.loaded_dataset = [t for t in dataset]
+        raise Exception("To be implemented by subclass")
 
     def getitem(self, idx: int):
         return self.generate_query(idx)
 
     def __len__(self):
-        return len(self.loaded_dataset)
+        raise Exception("To be implemented by subclass")
 
     def generate_query(self, idx: int) -> Tuple[List[dict], List[bytes]]:
         """
--- a/aperturedb/TensorFlowData.py
+++ b/aperturedb/TensorFlowData.py
@@ -3,7 +3,7 @@ from aperturedb.Subscriptable import Subscriptable
 import tensorflow as tf
 
 
-class TensorflowData(Subscriptable):
+class TensorFlowData(Subscriptable):
     """
     **Class to wrap around a Dataset retrieved from [Tensorflow datasets](https://www.tensorflow.org/datasets)**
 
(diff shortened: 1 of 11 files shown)


aperturedb/TensorFlowData.py
```diff
@@ -0,0 +1,45 @@
...
+    This class should be subclassed with a specific (custom) implementation of generate_query(),
+    and __init__ should be called with the dataset to be wrapped.
+    :::
+
+    Example subclass: [Cifar10DataTensorflow](https://github.com/aperture-data/aperturedb-python/blob/develop/examples/Cifar10DataTensorflow.py)
+
+    """
+
+    def __init__(self, dataset: tf.data.Dataset) -> None:
+        self.loaded_dataset = [t for t in dataset]
```
bovlb, line 24: Do we necessarily want to listify an arbitrarily large dataset here?

examples/Cifar10DataTensorFlow.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    def __init__(self):
+        (x_train, y_train), (x_val, y_val) = tf.keras.datasets.cifar10.load_data()
+        x = tf.concat([x_train, x_val], axis=0)
+        y = tf.concat([tf.squeeze(y_train), tf.squeeze(y_val)], axis=0)
+        ds = tf.data.Dataset.from_tensor_slices((x, y))
+        super().__init__(ds)
+
+    def __image__from__cifar10__(self, arr):
```
bovlb, line 21: Why is this a dunder-method?

examples/Cifar10DataTensorFlow.py
```diff
@@ -0,0 +1,39 @@
...
+        x, y = self.loaded_dataset[idx]
+        img_ref = (idx % 99998) + 1
+        q = [{
+            "AddImage": {
+                "_ref": img_ref
+            }
+        }]
+        q[0]["AddImage"]["properties"] = {
+            "label": str(y.numpy()),
+            "stage": "train" if idx < 50000 else "val"
```
bovlb, line 36: CIFAR-10 is documented as having 50,000 training images and 10,000 test images, but why don't we get this magic number from `x_train` in `__init__`?  And why are we labelling test images as `val`?

examples/CocoDataPyTorch.py
```diff
@@ -11,7 +11,7 @@ def image_to_byte_array(image: Image) -> bytes:
     return imgByteArr
 
 
-class CocoDataPytorch(PyTorchData):
+class CocoDataPyTorch(PyTorchData):
```
bovlb, line 14: This file and class are mentioned elsewhere.  Should those be fixed?

aperturedb/TensorFlowData.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def getitem(self, idx: int):
+        return self.generate_query(idx)
+
+    def __len__(self):
+        return len(self.loaded_dataset)
+
+    def generate_query(self, idx: int) -> Tuple[List[dict], List[bytes]]:
+        """
+        **Takes information from one atomic record from the Data and converts it to Query for apertureDB**
```
vishakha041, line 34:
```suggestion
        **Takes information from one atomic record from the Data and converts it to Query for ApertureDB**
```

aperturedb/TensorFlowData.py
gsaluja9: yes, this can be removed. will do so.

examples/Cifar10DataTensorFlow.py
gsaluja9: no need. Infact it should not even be a class method. This has been defined at other places too. I will consolidate and put it in a lib.

examples/Cifar10DataTensorFlow.py
gsaluja9: This was just following the guideline in this prior notebook that was built. I think it might be better to stick to the conventions of the dataset. will change the notebook, instead of this class. Also, will drop the magic number.

examples/CocoDataPyTorch.py
gsaluja9: OK I 'll do a find.

bovlb (approved):
LGTM

PR #332, opened 2023-10-26
Build and push docker image with aperturedb package

ali-asadpoor:


docker/release/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ubuntu:22.04
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt update && apt upgrade -y && apt install -y --no-install-recommends \
+    python3-dev \
+    python3-pip
+RUN pip3 install aperturedb
```
gsaluja9, line 6: Minor :
This assumes that immediately after twine upload is complete, pip will get the latest version from pypi. If this ends up pulling a version which is one before, we'll not know . we could set an env variable here, like we do for ADB_TAG and do a pip install aperturedb==$VER ?

gsaluja9: There's also a chance that the images that build on top of this might use https://github.com/aperture-data/aperturedb-python/blob/develop/aperturedb/PyTorchDataset.py, or some transformer which is still in PR. (I can think of benchmarks). If that is the case, it might be better to install aperturedb[complete]==$VER

ali-asadpoor: Passed the version.

gsaluja9 (approved):

PR #344, opened 2023-11-08
Updating the _ref values in a flattened batch.

gsaluja9:


aperturedb/BlobDataCSV.py
```diff
@@ -66,7 +66,6 @@ def get_indices(self):
         }
 
     def getitem(self, idx):
-        idx = self.df.index.start + idx
         filename = os.path.join(self.relative_path_prefix,
                                 self.df.loc[idx, BLOB_PATH])
```
drewaogle, line 70: The removed line is setting `idx`, but the new line 70 is still referencing `idx`

aperturedb/ParallelQuery.py
```diff
@@ -138,10 +138,34 @@ def __init__(self, db, dry_run=False):
...
+                    values["video_ref"] = updates[values["video_ref"]]
+                if "is_connected_to" in values and "_ref" in values["is_connected_to"]:
+                    values["is_connected_to"]["_ref"] = updates[values["is_connected_to"]["_ref"]]
+                if "src" in values:
+                    values["src"] = updates[values["src"]]
+                if "dst" in values:
+                    values["dst"] = updates[values["dst"]]
+            return batched_commands
+
+        q = update_refs([cmd for query in data for cmd in query[0]])
```
drewaogle, line 168: This looks good - is there any reason to keep the "video" and "image" since PQ is just doing matching at this point?

gsaluja9: not sure if I understand this correctly, but this would serve the case where getitem generates a transaction with both video and image, and entities with some connected to video, and others to images.

drewaogle: right - should we just change "video" and 'image" into something more generic?

aperturedb/BlobDataCSV.py
luisremis: idx is a param of the method.

luisremis (approved):

aperturedb/BlobDataCSV.py
drewaogle: right, but I would assume there was a reason idx was set there to begin with?

PR #345, opened 2023-11-08
Enabling axis on the plotlib show

gsaluja9:
Changing the notebook helper to show similar output. To show that native queries and Images class can display similar looking images.

### DIFF
--- a/aperturedb/Images.py
+++ b/aperturedb/Images.py
@@ -45,17 +45,30 @@ class Images(Entities):
     def inspect(self, use_thumbnails=True) -> Union[Tuple[widgets.IntSlider, widgets.Output], DataFrame]:
         df = super(Images, self).inspect()
         if use_thumbnails:
+            def image_base64(im):
+                with BytesIO() as buffer:
+                    im.save(buffer, 'jpeg')
+                    return base64.b64encode(buffer.getvalue()).decode()
+
+            def get_formatters(width):
+                def image_formatter(im):
+                    return f'<img width={width} style="max-width: 400px" src="data:image/jpeg;base64,{image_base64(im)}" >'
+
+                return image_formatter
+
             sizer = widgets.IntSlider(min=1, max=400, value=100)
             op = widgets.Output()
 
-            def widget_interaction(c):
-                def image_base64(im):
-                    with BytesIO() as buffer:
-                        im.save(buffer, 'jpeg')
-                        return base64.b64encode(buffer.getvalue()).decode()
+            with op:
+                image_formatter = get_formatters(sizer.value)
+                display(HTML("<div style='max-width: 100%; overflow: auto;'>" +
+                             df.to_html(
+                                 formatters={'thumbnail': image_formatter}, escape=False)
+                             + "</div>")
+                        )
 
-                def image_formatter(im):
-                    return f'<img width={c["new"]} style="max-width: 400px" src="data:image/jpeg;base64,{image_base64(im)}" >'
+            def widget_interaction(c):
+                image_formatter = get_formatters(c['new'])
                 with op:
                     op.clear_output()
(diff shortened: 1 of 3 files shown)


aperturedb/Images.py
```diff
@@ -45,17 +45,30 @@ class Images(Entities):
...
+                image_formatter = get_formatters(sizer.value)
+                display(HTML("<div style='max-width: 100%; overflow: auto;'>" +
+                             df.to_html(
+                                 formatters={'thumbnail': image_formatter}, escape=False)
+                             + "</div>")
+                        )
 
-                def image_formatter(im):
-                    return f'<img width={c["new"]} style="max-width: 400px" src="data:image/jpeg;base64,{image_base64(im)}" >'
+            def widget_interaction(c):
```
drewaogle, line 70: it's a little tricky to figure out what changed here - it looks like image_formatter became a function that is returned by get_formatters rather than one passed directly? was the img width not updating?

aperturedb/Images.py
```diff
@@ -110,6 +121,10 @@ def __init__(self, db, batch_size=100, response=None, **kwargs):
         self.bbox_label_prop = "_label"
         self.get_image = True
 
+        if response is not None:
+            self.images_ids = list(
+                map(lambda x: x[self.img_id_prop], response))
+
```
drewaogle, line 127: it's hard to tell what is gong on here - what exactly is response in this context? I assume we are getting the image_ids at construction if a response is passed in - so it this allowing a quick construct of a Images with an already in-hand response?

aperturedb/NotebookHelpers.py
```diff
@@ -53,12 +53,29 @@ def check_folder(folder):
...
+    This method was used by ApertureDB to display images, but it is not recommended anymore.
+    It will not draw annotation.
+
+    Instead, when using with JSON queries, make a instance of the Images class and call the display method.
+    :
+
+    ```python
+    from aperturedb.Images import Images
+    result, response, blobs = execute_batch([{"FindImage":{"uniqueids": True}}], [], db)
+    wrapper = Images(db, response=response[0]["FindImage"]["entities"])
```
drewaogle, line 68: ok, I got it - so that is what response is for.

But using this call won't display annotations either no?

drewaogle (commented):
just had some questions!

aperturedb/Images.py
gsaluja9: Yes, here is an [example](https://github.com/aperture-data/demos/blob/badf7f0d530b06321ac2f287392c5cfdab59c095/ml_notebooks/ROI-based%20Queries.ipynb) that would be useful to understand.

aperturedb/NotebookHelpers.py
gsaluja9: Images is pretty lazy loaded. Only if a subsequent calls to display has show_bboxes, or show_polygons, additional queries would be made. Even blobs happen lazily.

aperturedb/Images.py
gsaluja9: We've had a smallish widget for notebook . Its used in this [example](https://github.com/aperture-data/demos/blob/badf7f0d530b06321ac2f287392c5cfdab59c095/ml_notebooks/Vector-based%20Queries.ipynb). Refer to end of 6'th cell.

It had a problem on first render. This addresses it.

gsaluja9:
> just had some questions!

Responded. Better would be to watch me demo this. After standup .

aperturedb/NotebookHelpers.py
drewaogle: ok, cool, so more flexibility to pull in from a more complex query that you'd done by hand.

drewaogle (approved):
lgtm

PR #348, opened 2023-11-15
Do not raise an exception if BBox is not attached.

gsaluja9:
Continue processing for the rest of images.


vishakha041 (approved):
Looks good. I didn't run the test but seems to be setting all the empty/nonempty indexes

drewaogle (approved):
lgtm

PR #354, opened 2023-12-06
Remove docker pull to have a clean build without cache

ali-asadpoor:
The apt packages in `dependencies` image and any other image created from it are very old. Reusing old layers prevent docker from running `apt upgrade` again. This PR creates a cronjob to create `dependencies` image without any cache.

`build_images.sh` is not used in any part of repository and is removed.


ci.sh
```diff
@@ -166,8 +166,7 @@ build_complete(){
 build_notebook_dependencies_image(){
     DEPS_IMAGE=$DOCKER_REPOSITORY/aperturedb-notebook:dependencies
     echo "Building image $DEPS_IMAGE"
-    docker pull $DEPS_IMAGE
-    docker build -t $DEPS_IMAGE --cache-from $DEPS_IMAGE -f docker/dependencies/Dockerfile .
+    docker build -t $DEPS_IMAGE --no-cache -f docker/dependencies/Dockerfile .
```
luisremis, line 169: per our discussion, I think we will remove this line.

luisremis (approved):

.github/workflows/pr.yaml
```diff
@@ -35,61 +35,5 @@ jobs:
         AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         GCP_SERVICE_ACCOUNT_KEY: ${{ secrets.GCP_SERVICE_ACCOUNT_KEY }}
-      run: 'EXCLUDE_BUILD_COMPLETE=1 EXCLUDE_DEPLOY=1 NO_PUSH=1 BRANCH_NAME=${{ github.event.pull_request.head.ref }} TARGET_BRANCH_NAME=${{ github.event.pull_request.base.ref }} ./ci.sh'
+      run: RUN_TESTS=true NO_PUSH=true BRANCH_NAME=${{ github.event.pull_request.head.ref }} TARGET_BRANCH_NAME=${{ github.event.pull_request.base.ref }} ./ci.sh
       shell: bash
-
-  # run_test_conda_gpu:
```
luisremis, line 41: I think this code is commented but the intention is to fix the issue and put the tests back.
@gsaluja9 will know.

luisremis: I created an issue: https://github.com/aperture-data/aperturedb-python/issues/356

PR #359, opened 2023-12-15
Feature/fix streaming

gsaluja9:
Introducing a new functionality in ingest using adb, called 

1. --debug: just dumps the json and blobs for the samples to be inspected for correctness.
2. --sample_count: N , only generates queries corresponding to subset (of length N).

This also fixes a problem in updating references in generate batch, which was due to a missing check for "connect" in the batch.


aperturedb/ImageDataCSV.py
```diff
@@ -281,7 +281,7 @@ def getitem(self, idx):
         if self.format_given:
             custom_fields["format"] = self.df.loc[idx, IMG_FORMAT]
         ai = self._basic_command(idx, custom_fields)
-        ai[self.command]["_ref"] = (idx % 99998) + 1
```
drewaogle, line 284: if we're setting to 1 isn't it more appropriate just to leave it out?

gsaluja9: The expectation of update refs is that it needs a valid out come from getitem. So there's a check for ref to be available.

This is trivial scenario with just 1 image being added .
Consider a scenario where 1 command = 2 add images and 1 add bounding box and 1 add descriptor to the second image.

The parallel query should not assume any _ref if not there. This just makes it explicit.

drewaogle: I'm just worried that this could be confusing - can we change the check to run only when we expect a useful ref, and then we can disable it here?

It would be safer for something to break looking for a ref rather than a query run incorrectly.

gsaluja9: I had a look at it. Implicit refs would be tricky to do as generate batch does not know the item bondary from any getitem calls. I am going to make a comment on imagedataCSV so that this can be kept in mind.

gsaluja9: As per our call, I am merging this, as it adreses a flaky CI behavior. We can discuss this more if required.

PR #371, opened 2024-01-18
Fix for response_handler

luisremis:
Currently, response_handler is not being passed the right blobs corresponding to the response. This Branch has a workaround that will pass the right blobs under the special condition that we have BATCH_SIZE=1, but there is a need for a more general fix. 

This branch is needed for `nsfw_detector` to work: https://github.com/aperture-data/demos/pull/148

### DIFF
--- a/aperturedb/ParallelQuery.py
+++ b/aperturedb/ParallelQuery.py
@@ -14,7 +14,7 @@ logger = logging.getLogger(__name__)
 
 
 def execute_batch(q, blobs, db, success_statuses: list[int] = [0],
-                  response_handler: Callable = None, commands_per_query: int = 1, blobs_per_query: int = 1,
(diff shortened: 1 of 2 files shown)


gsaluja9:
> Currently, response_handler is not being passed the right blobs corresponding to the response. This Branch has a workaround that will pass the right blobs under the special condition that we have BATCH_SIZE=1, but there is a need for a more general fix.
> 
> This branch is needed for `nsfw_detector` to work: [aperture-data/demos#148](https://github.com/aperture-data/demos/pull/148)

This has been addresses and a test case added (with mock DB) added to it.

gsaluja9:
With limit and offset defined as 128, I was able to set a batch size of 2 and got a blobs array of 256, and 2 invocations of response handlers, with 128 blobs each.

gsaluja9:
Do not delete this branch for cleanup till demos is updated.

PR #377, opened 2024-02-06
Feature/client operations

gsaluja9:
To unblock demos. This introduces client operations so that annotations on images can be resolved to match the server side operations on images.

Currently implements resize and rotate.


drewaogle (approved):
lgtm

PR #383, opened 2024-02-15
ability to specify operations in python Query.

gsaluja9:
This is to enable the operations along with python based query API we have in demos.


aperturedb/Operations.py
```diff
@@ -31,3 +35,39 @@ def rotate(self, angle, resize=False):
...
+            "x": x,
+            "y": y,
+            "width": width,
+            "height": height,
+        }
+
+        self.operations_arr.append(op)
+        return self
+
+    def interval(self, start: int, stop: int, step: int) -> Operations:
```
luisremis, line 63: this operation is only available for videos or clips, not sure how this will impact.

luisremis (approved):
thanks!

aperturedb/Operations.py
gsaluja9: I am guessing server will send back an appropriate response if a invalid combo land up.

PR #385, opened 2024-02-16
Handle exceptions in retry loop

ali-asadpoor:
Some exceptions raised in http retry loop prevented retry to continue. I added exception handling to let retry finish its work.

### DIFF
--- a/aperturedb/ConnectorRest.py
+++ b/aperturedb/ConnectorRest.py
@@ -132,11 +132,11 @@ class ConnectorRest(Connector):
                         b) for b in json_response['blobs']]
                     self.last_response  = json_response["json"]
                     break
-            except Exception as e:
-                response_text = response.text[:1000] if hasattr(
-                    response, "text") else ""
                 logger.error(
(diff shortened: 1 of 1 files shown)


aperturedb/ConnectorRest.py
```diff
@@ -119,21 +119,25 @@ def _query(self, query, blob_array = [], try_resume=True):
...
+                    json_response       = json.loads(response.text)
+                    import base64
+                    response_blob_array = [base64.b64decode(
+                        b) for b in json_response['blobs']]
+                    self.last_response  = json_response["json"]
+                    break
+            except Exception as e:
+                response_text = response.text[:1000] if hasattr(
+                    response, "text") else ""
+                logger.error(
```
gsaluja9, line 138: This logger will the case when there is no exception but http status code is no OK.

ali-asadpoor: Added a separate log for each case.

gsaluja9 (approved):

PR #392, opened 2024-02-29
Document Connector.create_new_connection

bovlb:
This method is crucial for writing multi-threaded applications.  A customer reported a problem this morning, and we are about to recommend that they fix it using this (previously-undocumented) method.


gsaluja9 (approved):
thanks

PR #403, opened 2024-03-20
First pass at schema visualization

bovlb:
This adds a `visualize_schema` method to `Utils` and a corresponding `visualize-schema` command to the CLI.

Simply call
```
adb utils visualize-schema
```
or 
```
adb utils visualize-schema --filename=schema --format=png
```

Relies on graphviz installation.
schema

PR #409, opened 2024-04-03
Change schema visualization colours to use website palette

bovlb:
Colours changed and formatting tweaked.  

See https://github.com/aperture-data/aperturedb-python/pull/403 for before example.

After:
image


vishakha041 (approved):
We can always use white for the letters. but this is good to go

PR #412, opened 2024-04-20
Remove unused `get_uniqueids` and `get_images_uniqueids`

luisremis:
Newer versions of [aperturedb API will prevent the use of internal class names](https://github.com/aperture-data/athena/pull/1490) (i.e., `_Image`, `_Descriptor`, etc) on `FindEntity` commands. `get_uniqueids` and `get_images_uniqueids` rely on that to work correctly.

Currently, there seem to be no applications using `get_uniqueids` and `get_images_uniqueids`, nor internal tests cases or classes that use any them.


gsaluja9:
- [ ] The docs build would need to be re triggered.
- [ ] This is being used in one of the customer app repo. Would need a removal.

Other than that, changes are good to go.

gsaluja9 (approved):
lgtm

gsaluja9:
This also affects various calls to count_entities in utils to @luisremis , @drewaogle

drewaogle:
Sounds good. I don't use any of these directly, iirc.

PR #413, opened 2024-04-22
Takes a ImageDataCSV as input, along with a transformer, to create a DescriptorDataCSV and npy file to be used with adb.

gsaluja9:


aperturedb/cli/ingest.py
```diff
@@ -214,3 +218,62 @@ def from_csv(filepath: Annotated[str, typer.Argument(
...
+                               adb_data_source=f"{os.path.basename(input_file)}")
+    filename = f"{os.path.basename(input_file)}_{all_transformers[-1]}"
+    metadata = []
+    connection = []
+    embeddings = []
+
+    for i in tqdm(range(data.sample_count)):
+        d = data[i]
+        embeddings.append(np.frombuffer(d[1][1], dtype=np.float32))
+        metadata.append({
```
luisremis, line 261: If speed becomes (or is) an issue, generating [pandas df columns directly](https://github.com/aperture-data/demos/blob/master/coco/generate_coco_csv.py#L26) may be faster.

luisremis (approved):

aperturedb/cli/ingest.py
gsaluja9: There is speed thing, but it's because of model's inference. For now, this is mostly a util to avoid run time inference cost.

PR #415, opened 2024-04-25
Better ids for connections between images and descriptors

gsaluja9:
This would have failed for datasets with no id property.
Tested when running for celebA DS with 2 descriptor sets.


luisremis (approved):
thanks!

PR #416, opened 2024-04-25
Do not log first connection attempt.

gsaluja9:
This would result in a warn log when trying to connect for the first time, which is not really a case for warning.

### DIFF
--- a/aperturedb/Connector.py
+++ b/aperturedb/Connector.py
@@ -366,7 +366,7 @@ class Connector(object):
         data = query_msg.SerializeToString()
 
         # this is for session refresh attempts
-        tries = 0
+        tries = -1
         while tries < self.config.retry_max_attempts:
             try:
                 if self._send_msg(data):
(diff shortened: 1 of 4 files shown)


aperturedb/Connector.py
```diff
@@ -366,7 +366,7 @@ def _query(self, query, blob_array = [], try_resume=True):
         data = query_msg.SerializeToString()
 
         # this is for session refresh attempts
-        tries = 0
+        tries = -1
```
luisremis, line 369: this means there will be self.config.retry_max_attempts + 1 retries. 

I think what you want is to keep `tries = 0` and move line 408 to 414.

luisremis (approved):

aperturedb/Connector.py
gsaluja9: my dilemma is on the variable name. The first is technically not a retry, but this is the only place for retry mechanism. 

Let me think more . 

I though about moving 408 to 414 but that'd be 3 connection attempts, 2 of which are actual retries.

gsaluja9: we could call it connection attempts instead of retry and be done with it. Thoughts?

PR #419, opened 2024-05-03
Do not assume an subscriptable of errors when commands fail at aperturedb

gsaluja9:
This used to result in errors on the line of :
unhashable type: 'slice', which was not a useful (underlying) bit to be shown to the user.

For example, with a return code 3, the query or transaction does not return a list which was the earlier expectation in calculating the actual stats.

### DIFF
--- a/aperturedb/ParallelQuery.py
+++ b/aperturedb/ParallelQuery.py
@@ -58,18 +58,19 @@ def execute_batch(q: Commands, blobs: Blobs, db: Connector,
                 blobs_end = blobs_start + blobs_per_query
 
                 b_count = 0
-                for req, resp in zip(q[start:end], r[start:end]):
-                    for k in req:
-                        # Ref to https://docs.aperturedata.io/query_language/Reference/shared_command_parameters/blobs
-                        blobs_where_default_true = \
-                            k in ["FindImage", "FindBlob", "FindVideo"] and (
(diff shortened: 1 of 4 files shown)


aperturedb/ParallelQuery.py
```diff
@@ -274,8 +274,11 @@ def filter_per_group(group):
                     [v['status'] == 2 for i in r for k, v in filter_per_group(i)])
                 sq = 0
                 for i in range(0, len(r), self.commands_per_query):
-                    if all([v['status'] == 0 for j in r[i:i + self.commands_per_query] for k, v in filter_per_group(j)]):
-                        sq += 1
+                    # Some errors stop the whole query from being executed
+                    # https://docs.aperturedata.io/query_language/Overview/Responses#return-status
+                    if issubclass(type(r), list):
+                        if all([v['status'] == 0 for j in r[i:i + self.commands_per_query] for k, v in filter_per_group(j)]):
+                            sq += 1
```
drewaogle, line 281: Should we so something for the else here

luisremis (approved):
thanks!

aperturedb/ParallelQuery.py
gsaluja9: I have added tests. I think some cleanup is needed on what we compute as actual stats, but this PR does not address those concerns.

PR #420, opened 2024-05-04
Descriptors wrapper

bovlb:
Simple descriptors wrapper that just provides `find_similar`and `find_similar_mmr` to use in langchain.

### DIFF
--- a/aperturedb/Descriptors.py
+++ b/aperturedb/Descriptors.py
@@ -3,6 +3,7 @@ import logging
 import numpy as np
 
 from aperturedb.Entities import Entities
+from aperturedb.ParallelQuery import execute_batch
 
 logger = logging.getLogger(__name__)
 
@@ -14,9 +15,8 @@ class Descriptors(Entities):
 
     db_object = "_Descriptor"
 
-    def __init__(self, db, response=None):
-        super().__init__(db, response)
-        self.db_connector = db
+    def __init__(self, db):
+        super().__init__(db)
 
     def find_similar(self,
                      set: str,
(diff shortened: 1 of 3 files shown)


aperturedb/Descriptors.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class Descriptors(Entities):
+    """
+    Python wrapper for ApertureDB Descriptors API.
+    """
+
+    db_object = "_Descriptor"
+
+    def __init__(self, db, response=None):
```
gsaluja9, line 17: The response, in Images is to facilitate a easy wrapper creation when handling a response from a JSON query. The helper methods become easy to use then. This is obviously not interface specific, but will be a good to consider.

bovlb: OK.  I think that would have worked, but I don't think it's as useful for descriptors, so I took it out.

aperturedb/Descriptors.py
```diff
@@ -0,0 +1,150 @@
...
+        }}
+
+        if constraints is not None:
+            command["FindDescriptor"]["constraints"] = constraints.constraints
+
+        query = [command]
+        logger.debug(query)
+        blobs_in = [np.array(vector, dtype=np.float32).tobytes()]
+
+        response, blobs_out = self.db_connector.query(query, blobs_in)
```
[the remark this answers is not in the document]
bovlb: Done

gsaluja9 (approved):
Thanks. Let's merge it.

PR #430, opened 2024-05-21
Add warning for message size larger than 256MB

luisremis:
Related: https://github.com/aperture-data/athena/issues/1525

### DIFF
--- a/aperturedb/Connector.py
+++ b/aperturedb/Connector.py
@@ -182,6 +182,11 @@ class Connector(object):
             self.connected = False
 
     def _send_msg(self, data):
+        # aperturedb's param ADB_MAX_CONNECTION_MESSAGE_SIZE_MB = 256 by default
+        if len(data) > (256 * 2**20):
+            logger.warning(
+                "Message sent is larger than default for ApertureDB Server. Server may disconnect.")
+
         sent_len = struct.pack('@I', len(data))  # send size first
         x = self.conn.send(sent_len + data)
         return x == len(data) + 4


aperturedb/Connector.py
```diff
@@ -183,6 +183,11 @@ def __del__(self):
 
     def _send_msg(self, data):
         sent_len = struct.pack('@I', len(data))  # send size first
+        
+        # aperturedb's param ADB_MAX_CONNECTION_MESSAGE_SIZE_MB = 256 by default
+        if send_len > (256 * 2**20):
+            logger.warning("Message sent is larger than default for ApertureDB Server. Server may disconnect.")
+
```
luisremis, line 190:
```suggestion
    def _send_msg(self, data):
        # aperturedb's param ADB_MAX_CONNECTION_MESSAGE_SIZE_MB = 256 by default
        if len(data) > (256 * 2**20):
            logger.warning("Message sent is larger than default for ApertureDB Server. Server may disconnect.")
        
        sent_len = struct.pack('@I', len(data))  # send size first
```

aperturedb/Connector.py
```diff
@@ -182,6 +182,10 @@ def __del__(self):
             self.connected = False
 
     def _send_msg(self, data):
+        # aperturedb's param ADB_MAX_CONNECTION_MESSAGE_SIZE_MB = 256 by default
+        if len(data) > (256 * 2**20):
+            logger.warning("Message sent is larger than default for ApertureDB Server. Server may disconnect.")
```
luisremis, line 187:
```suggestion
            logger.warning(
                   "Message sent is larger than default for ApertureDB Server. Server may disconnect.")
```

aperturedb/Connector.py
```diff
@@ -182,6 +182,11 @@ def __del__(self):
             self.connected = False
 
     def _send_msg(self, data):
+        # aperturedb's param ADB_MAX_CONNECTION_MESSAGE_SIZE_MB = 256 by default
+        if len(data) > (256 * 2**20):
+            logger.warning(
+                   "Message sent is larger than default for ApertureDB Server. Server may disconnect.")
```
luisremis, line 188:
```suggestion
                "Message sent is larger than default for ApertureDB Server. Server may disconnect.")
```

aperturedb/Connector.py
```diff
@@ -182,6 +182,11 @@ def __del__(self):
             self.connected = False
 
     def _send_msg(self, data):
+        # aperturedb's param ADB_MAX_CONNECTION_MESSAGE_SIZE_MB = 256 by default
+        if len(data) > (256 * 2**20):
+            logger.warning(
+                "Message sent is larger than default for ApertureDB Server. Server may disconnect.")
```
gsaluja9, line 188: Curious to know why the word is "may"? I thought that it was a hard limit.
Should we just fail the query, and write a scenario in tests when this is triggered?

luisremis: it "may" disconnect because the server may use a different configuration than the default. If the server uses a different configuration than the default, the server will disconnect if the message size is larger than the set value (which can be more or less than 256MB).

Until we do a new client-server protocol, there is no way to communicate the limit to the client without breaking compatibility, that's why we are holding off the change.

gsaluja9 (approved):

PR #431, opened 2024-05-23
Add response handler support to ParallelQuerySet

drewaogle:
Move response_handler generation into ParallelQuery, simplyfing execute_batch.
Adds response_handler for ParallelQuerySet, defined as set_number,query,query_blobs,response,response_blobs.

### DIFF
--- a/aperturedb/ParallelQuery.py
+++ b/aperturedb/ParallelQuery.py
@@ -126,14 +126,9 @@ class ParallelQuery(Parallelizer.Parallelizer):
             if issubclass(type(response), list):
                 for req, resp in zip(query[start:end], response[start:end]):
                     for k in req:
-                        # Ref to https://docs.aperturedata.io/query_language/Reference/shared_command_parameters/blobs
-                        blobs_where_default_true = \
-                            k in ["FindImage", "FindBlob", "FindVideo"] and (
-                                "blobs" not in req[k] or req[k]["blobs"])
-                        blobs_where_default_false = \
-                            k in [
-                                "FindDescriptor", "FindBoundingBox"] and "blobs" in req[k] and req[k]["blobs"]
-                        if blobs_where_default_true or blobs_where_default_false:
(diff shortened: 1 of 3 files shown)


aperturedb/ParallelQuery.py
```diff
@@ -140,6 +110,44 @@ def setSuccessStatus(cls, statuses: list[int]):
...
+            end = start + commands_per_query
+            blobs_start = i * blobs_per_query
+            blobs_end = blobs_start + blobs_per_query
+
+            b_count = 0
+            if issubclass(type(response), list):
+                for req, resp in zip(query[start:end], response[start:end]):
+                    for k in req:
+                        # Ref to https://docs.aperturedata.io/query_language/Reference/shared_command_parameters/blobs
+                        blobs_where_default_true = \
```
gsaluja9, line 130: This is not correct anymore. blobs=False is default behavior

gsaluja9 (changes requested):
response blob counted needs to be changed to align with adb blobs=false for find*

aperturedb/ParallelQuery.py
drewaogle: fixed.

test/test_ResponseHandler.py
```diff
@@ -220,3 +304,46 @@ def test_response_blobs(self, db, utils, monkeypatch):
...
+        self.responses = {}
+        self.response_blobs = {}
+        generator = QGSetPersonAndImages(self.requests, self.responses,
+                                         max_commands, self.response_blobs)
+        querier = ParallelQuerySet(db)
+        querier.query(generator, batchsize=99,
+                      numthreads=31,
+                      stats=True)
+
+        def expected_blobs_for_resp(rh_responses):
```
gsaluja9, line 323: this can be removed as it is not being used

gsaluja9 (dismissed):

gsaluja9 (approved):

PR #439, opened 2024-06-06
Make a ML ready notebook.

gsaluja9:


bovlb (approved):
I'm not certain that we necessarily want CLIP in every notebook, but otherwise LGTM.

gsaluja9:
> I'm not certain that we necessarily want CLIP in every notebook, but otherwise LGTM.

I see why you'd be saying this (wrt semantic search), but the rationale is as follows:
This is a baseline setup that should be able to run all the features of adb (transformers, fuse stuff, etc.) Usage of this notebook would be advocated in Troubleshooting section of setting up the client. 

I'd love to discuss a more cleanly cut approach.

PR #440, opened 2024-06-12
fix some typos

bovlb:
Given that this is the first thing many clients see, I think this could use more polishing, but this is a start.  In particular, I don't think the "Running tests" section adequately describes what is required to get the tests running.


gsaluja9 (approved):

gsaluja9:
thanks

PR #441, opened 2024-06-13
Add descriptorset properties

bovlb:
Adding optional properties parameter to `Utils.add_descriptorset`.  This is useful for clients that want to record information about the descriptorset.  I plan to use it in the langchain integration.


gsaluja9 (approved):
looks good.

PR #445, opened 2024-06-17
Pin numpy to less than 2.

gsaluja9:
numpy 2.0 creates incompatibility, and was released on Jun 16.
Pinning this till a better fix is available.


drewaogle (dismissed):
looks like it should work!

drewaogle (approved):
all good now with quotes

PR #448, opened 2024-06-19
Add 'index' parameter for response_handler

drewaogle:
It's optional, we wrap in ParallelQuery at beginning of loop.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,4 +15,4 @@ repos:
       hooks:
       - id: autopep8
         exclude: _pb2.py$
-        args: ["--ignore", "E251,E241,E221,E402", "-i"]
+        args: ["--ignore", "E251,E241,E221,E402,E275", "-i"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,4 +15,4 @@ repos:
       hooks:
       - id: autopep8
         exclude: _pb2.py$
-        args: ["--ignore", "E251,E241,E221", "-i"]
+        args: ["--ignore", "E251,E241,E221,E402", "-i"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ ci:
     skip: []
     submodules: false
 repos:
-    - repo: https://github.com/hhato/autopep8
+    - repo: https://github.com/hhatto/autopep8
       rev: 8b75604
       hooks:
       - id: autopep8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,4 +15,4 @@ repos:
       hooks:
       - id: autopep8
         exclude: _pb2.py$
-        args: ["--ignore", "E251,E241,E221,E402,E275", "-i"]
+        args: ["--ignore", "E251,E241,E221,E402,E265,E275", "-i"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,8 +10,8 @@ ci:
     skip: []
     submodules: false
 repos:
-    - repo: https://github.com/pre-commit/mirrors-autopep8
-      rev: 898691a
+    - repo: https://github.com/hhato/autopep8
+      rev: 8b75604
       hooks:
       - id: autopep8
         exclude: _pb2.py$
--- a/aperturedb/ParallelQuery.py
+++ b/aperturedb/ParallelQuery.py
@@ -18,7 +18,7 @@ logger = logging.getLogger(__name__)
 def execute_batch(q: Commands, blobs: Blobs, db: Connector,
                   success_statuses: list[int] = [0],
                   response_handler: Optional[Callable] = None, commands_per_query: int = 1, blobs_per_query: int = 0,
-                  strict_response_validation: bool = False,cmd_index=None) -> Tuple[int, CommandResponses, Blobs]:
+                  strict_response_validation: bool = False, cmd_index=None) -> Tuple[int, CommandResponses, Blobs]:
     """
     Execute a batch of queries, doing useful logging around it.
     Calls the response handler if provided.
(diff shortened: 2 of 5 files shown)


test/test_ResponseHandler.py
```diff
@@ -339,3 +346,33 @@ def test_set_response(self, db, utils, max_commands, monkeypatch):
...
+                      numthreads=1,
+                      stats=True)
+        dist_by_ages = {}
+        # req and resp are dict mapping index number to query.
+        for reqk, respk in zip(self.requests.keys(), self.responses.keys()):
+            req=self.requests[reqk]
+            age_group = req[0]["FindEntity"]["constraints"]["age"][1]
+            # in QGPersonIndex, age_group[1] is created by taking index
+            # and multiplying it by cqp and 10.
+            assert reqk == ( 0 if age_group == 0 else age_group / cpq / 10 )
```
bovlb, line 378: Why is this not?
```suggestion
            assert reqk ==  age_group // cpq // 10 )
```

drewaogle: for some reason I was thinking denominator could be zero

aperturedb/ParallelQuery.py
```diff
@@ -257,6 +260,13 @@ def process_responses(requests, input_blobs, responses, output_blobs):
                 response_handler = self.generator.response_handler
             if hasattr(self.generator, "strict_response_validation") and isinstance(self.generator.strict_response_validation, bool):
                 strict_response_validation = self.generator.strict_response_validation
+
+            # if response_handler doesn't support index, just discard the index with a wrapper.
+            if response_handler is not None:
+                if not 'index' in inspect.signature(response_handler).parameters:
```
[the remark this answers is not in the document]
drewaogle: it wouldn't work. also if they happened to have a 4 query version and called the first parameter 'index' it would not wrap it.

Perhaps we just want to see if parameters is greater than 4?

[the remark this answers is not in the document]
bovlb: > Perhaps we just want to see if parameters is greater than 4?

That might be better.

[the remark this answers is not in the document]
drewaogle: done

bovlb (dismissed):
LGTM

gsaluja9 (dismissed):

.pre-commit-config.yaml
```diff
@@ -10,9 +10,9 @@ ci:
...
 repos:
-    - repo: https://github.com/pre-commit/mirrors-autopep8
-      rev: 898691a
+    - repo: https://github.com/hhatto/autopep8
+      rev: 8b75604
       hooks:
       - id: autopep8
         exclude: _pb2.py$
-        args: ["--ignore", "E251,E241,E221", "-i"]
+        args: ["--ignore", "E251,E241,E221,E402,E265,E275", "-i"]
```
bovlb, line 18: Now I'm intrigued.  Why do we need this growing list?

I believe this is the corresponding list: https://pycodestyle.pycqa.org/en/latest/intro.html

aperturedb/ParallelQuery.py
```diff
@@ -316,7 +333,8 @@ def worker(self, thid: int, generator, start: int, end: int):
             batch_end = min(batch_start + self.batchsize, end)
 
             try:
-                self.do_batch(db, generator[batch_start:batch_end])
+                self.do_batch(db, batch_start,
+                              generator[batch_start:batch_end])
```
bovlb, line 337: It's probably a bad idea to call this parameter "generator" if we're going to index it like this.

test/test_ResponseHandler.py
```diff
@@ -333,9 +349,54 @@ def test_set_response(self, db, utils, max_commands, monkeypatch):
...
+        self.cleanDB()
+        persons = EntityWithResponseDataCSV(
+            "./input/persons.adb.csv", self.requests, self.responses)
+        loader = ParallelLoader(db)
+        loader.ingest(persons, batchsize=99,
+                      numthreads=31,
+                      stats=True)
+        assert loader.error_counter == 0
+
+        dist = persons.df.groupby(persons.df.age // 10 * 10).count()
```
bovlb, line 370: Is this variable ever used?

test/test_ResponseHandler.py
```diff
@@ -333,9 +349,54 @@ def test_set_response(self, db, utils, max_commands, monkeypatch):
...
+        self.requests = {}
+        self.responses = {}
+        self.response_blobs = {}
+        generator = QGPersonsIndex(self.requests, self.responses,
+                                   cpq, self.response_blobs)
+        querier = ParallelQuery(db)
+        querier.query(generator, batchsize=100,
+                      numthreads=1,
+                      stats=True)
+        dist_by_ages = {}
```
bovlb, line 381: Is this variable ever used?

bovlb (approved):
LGTM

.pre-commit-config.yaml
gsaluja9: Should we decide on a list of exclusions. I believe this list was to support our existing style, but I think it should be OK to let it all be rescanned even if it leads to a big diff.

drewaogle: not sure; I generated this by trying to run the new autopep on the code. I used `pycodestyle` on the files that were failing that I hadn't changed to find out what it was complaining about.

I added ignores until all the currently existing files passed.

bovlb: Good answer.

Some projects have an autoformat feature where it will correct such errors automatically.

aperturedb/ParallelQuery.py
drewaogle: I assume you are talking in terms of a python generator - my assumption was that it was named this because it is a QueryGenerator ( or was initially )

test/test_ResponseHandler.py
drewaogle: no, removed. I changed how the code calculated.

test/test_ResponseHandler.py
drewaogle: same.

gsaluja9 (approved):

PR #449, opened 2024-06-25
Make connection property font color darker

bovlb:
I noticed during the Exthymic tech talk that the text for connection properties wasn't very readable.

Before:

image

After:

image


luisremis (approved):
thanks!

PR #450, opened 2024-06-26
Allow logging file to be controlled by env variable

drewaogle:


bovlb (approved):
LGTM

aperturedb/__init__.py
```diff
@@ -19,11 +21,23 @@
...
 
 # define file handler and set formatter
-error_file_handler = logging.FileHandler(
-    f"error.{datetime.datetime.now().isoformat()}.log", delay=True)
-error_file_handler.setFormatter(formatter)
-error_file_handler.setLevel(log_file_level)
-logger.addHandler(error_file_handler)
+error_file_name = "error.${now}.log"
+
+if "ADB_LOG_FILE" in os.environ:
```
gsaluja9, line 26: So this won't create a file if ADB_LOG_FILE is not set, and this would be most likely the case.
Why can't we have an option to explicitly disable (or change the path) of it, and fall back to this?

gsaluja9 (approved):

PR #451, opened 2024-07-09
protect from empty entities list

bovlb:


aperturedb/Descriptors.py
```diff
@@ -57,7 +61,7 @@ def find_similar(self,
         blobs_in = [np.array(vector, dtype=np.float32).tobytes()]
         _, response, blobs_out = execute_batch(query, blobs_in, self.db)
 
-        self.response = response[0]["FindDescriptor"]["entities"]
+        self.response = response[0]["FindDescriptor"].get("entities", [])
```
gsaluja9, line 64: This is nice. it leads to ambiguous messages in some helper functions.

gsaluja9 (approved):

PR #456, opened 2024-07-24
expand protobuf support to 4.x

drewaogle:


aperturedb/Connector.py
```diff
@@ -380,8 +380,8 @@ def _query(self, query, blob_array = [], try_resume=True):
                 if self._send_msg(data):
                     response = self._recv_msg()
                     if response is not None:
-                        querRes = queryMessage_pb2.queryMessage()
-                        querRes.ParseFromString(response)
+                        querRes = queryMessage.queryMessage()
+                        querRes.ParseFromString(bytes(response))
```
drewaogle, line 384: this was added because response is a `bytearray` in 4.x and ParseFromString was expecting a `bytes type object` which `bytearray` *is* but it wouldn't accept.

This may not be the optimal method.

drewaogle: Due to this: https://github.com/protocolbuffers/protobuf/issues/10774

aperturedb/Connector.py
```diff
@@ -380,8 +380,9 @@ def _query(self, query, blob_array = [], try_resume=True):
                 if self._send_msg(data):
                     response = self._recv_msg()
                     if response is not None:
-                        querRes = queryMessage_pb2.queryMessage()
-                        querRes.ParseFromString(response)
+                        querRes = queryMessage.queryMessage()
+                        querRes.ParseFromString(response if
```
gsaluja9, line 384: How about making a Parse and Build method in queryMessage, so that all the details of dealing with querymessage are contained in one place.

gsaluja9 (approved):
Thanks . this would help many users! Just a feedback to make it an interface

aperturedb/Connector.py
drewaogle: that seems like a good idea.

aperturedb/Connector.py
drewaogle: changed to use memoryview, should be more performant, some people seem to https://stackoverflow.com/questions/55772903/python-memoryview-slower-than-expected say it can be slower than `bytes` but perhaps that is due to it optimizing the copy out in that situation.

aperturedb/Connector.py
drewaogle: added.

bovlb:
Will we have any automated testing of a range of protobuf versions?

drewaogle:
I think that is a good idea, but is a bit more complex to decide what we should do.

bovlb (approved):
LGTM

One question about what we're testing.

drewaogle:
did testing on conversion:
```
Testing list(b[i:i+1000])
Size: 1000000,bytes(x) to  bytes, time: 0.04472851753234863
Size: 1000000,memoryview(x).tobytes() to  bytes, time: 0.04598498344421387
=====================
Testing list(b[i:])
Size: 1000000,bytes(x) to  bytes, time: 0.9096970558166504
Size: 1000000,memoryview(x).tobytes() to  bytes, time: 0.9073359966278076
=====================
Testing b[i:]
Size: 1000000,bytes(x) to  bytes, time: 0.05309891700744629
Size: 1000000,memoryview(x).tobytes() to  bytes, time: 0.05276131629943848
```

looks like for some ops converting bytesarray to bytes is slightly faster, but it is probably a wash

PR #460, opened 2024-08-10
Support named config in create_connector

bovlb:
The `adb config` does not easily support having multiple connectors with different configurations in the same application (except by messing around with the environment variables).  This change allows the user to specify an explicit configuration name.  I have also tidied up the code, guarded against missing data, and made the documentation clearer.


drewaogle (approved):
lgtm

Issue #467, opened 2024-08-23
SDK trial on Mac


vishakha041:
Drew’s tagged version worked for ApertureDB server.

1. Adb was quite slow weirdly. 3 sec to load. Config ls - looks like we load unnecessary packages for certain commands like config ls. After caching - 1.4s  --> PYTHONPROFILEIMPORTTIME=1 adb config ls
2. If you run adb ingest without config - it just hung.
3. We print admin password in stack trace
4. If you create a config with same name twice - it just replaces existing config. Not sure that’s intentional.
5. Adb ingest doesn’t work due to the port forwarding stuff. Adb UI connected.
6. Our error stack for python is so long it’s too much. Is ipython slowing stuff too much ? Hurting user experience

vishakha041:
(adb) --($:~/.../adb/slim-stanford-car-dataset)---$ adb utils execute summary
2024-08-23 11:13:31,493 : ERROR : aperturedb.Connector : 8653999936 : 330 : Error connecting to server: [Errno 54] Connection reset by peer [127.0.0.1:456 as admin using TCP with SSL=True]
Traceback (most recent call last):
  File "/Users/rcledat/mambaforge/envs/adb/lib/python3.10/site-packages/aperturedb/Connector.py", line 306, in _connect
    response = self._recv_msg()
  File "/Users/rcledat/mambaforge/envs/adb/lib/python3.10/site-packages/aperturedb/Connector.py", line 195, in _recv_msg
    recv_len = self.conn.recv(4)  # get message size
ConnectionResetError: [Errno 54] Connection reset by peer
Stack (most recent call last):
  File "/Users/rcledat/mambaforge/envs/adb/bin/adb", line 8, in <module>
    sys.exit(app())
  File "/Users/rcledat/mambaforge/envs/adb/lib/python3.10/site-packages/typer/main.py", line 309, in __call__
    return get_command(self)(*args, **kwargs)
  File "/Users/rcledat/.local/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/Users/rcledat/mambaforge/envs/adb/lib/python3.10/site-packages/typer/core.py", line 723, in main
    return _main(
  File "/Users/rcledat/mambaforge/envs/adb/lib/python3.10/site-packages/typer/core.py", line 193, in _main
    rv = self.invoke(ctx)
[...]

PR #468, opened 2024-08-23
Reordered imports to lazily import them .Makes CLI snappier.

gsaluja9:
Addresses part of #467 .
adb config * run under 0.5 seconds.
Screenshot 2024-08-27 at 4 58 53 PM


bovlb (dismissed):
LGTM

In Utils.py, consider deferring the import of Graphviz.

PR #470, opened 2024-08-26
Misc usability issues.

gsaluja9:


aperturedb/Utils.py
```diff
@@ -120,7 +120,7 @@ def __init__(self, connector: Connector, verbose=False):
         self.verbose = verbose
 
     def __repr__(self):
-        return self.status()
+        return f"{id(self)=}"
```
drewaogle, line 123: We will need to bump min python to 3.8 to use the {var=}

aperturedb/Connector.py
```diff
@@ -167,11 +167,7 @@ def authenticate(self, shared_data, user, password, token):
         if not self.authenticated:
             if shared_data.session is None:
                 self.shared_data.lock = Lock()
-                try:
-                    self._authenticate(user, password, token)
-                except Exception as e:
-                    raise UnauthenticatedException(
-                        "Authentication failed:", str(e))
+                self._authenticate(user, password, token)
```
drewaogle, line 170: This means that authenticate will just return on a failed authenticate?

gsaluja9: All the retry, with connect and refresh had been moved inside common function _query, which takes care of the retry mechanism. This exception need not be handled anywhere, and the one raised in _query will be the one to catch and validate.

aperturedb/Utils.py
gsaluja9: oh. we do not need the = . It's just a habbit. I will remove that

drewaogle (approved):
lgtm

PR #471, opened 2024-08-28
Removing non essential imports, timeouts, and overwrite confirmation

gsaluja9:
using ssl from adb config.
fixes some points of #467

### DIFF
--- a/aperturedb/Connector.py
+++ b/aperturedb/Connector.py
@@ -276,6 +276,10 @@ class Connector(object):
     def _connect(self):
 
         self.conn = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        # Setting a reasonable timeout value for the connection
+        # default is None which leads to blocking call till system does something.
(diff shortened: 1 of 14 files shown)


aperturedb/Connector.py
```diff
@@ -275,6 +276,10 @@ def _refresh_token(self):
     def _connect(self):
 
         self.conn = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        # Setting a reasonable timeout value for the connection
+        # default is None which leads to blocking call till system does something.
+        # https://docs.python.org/3/library/socket.html#timeouts-and-the-connect-method
+        self.conn.settimeout(1)
```
luisremis, line 282: I think 1 second may be too little, but it is a quick change we can do after, if we see this timeout kicks in too often.

luisremis (approved):
thanks!

aperturedb/Connector.py
gsaluja9: yea, if need be, this can put in config, later too. we will know in some ingestion runs, before making a release.

PR #476, opened 2024-09-09
Reverting the timeout change.

gsaluja9:
This had an unwanted sideeffect in longer reads.

### DIFF
--- a/aperturedb/Connector.py
+++ b/aperturedb/Connector.py
@@ -274,12 +274,7 @@ class Connector(object):
             raise UnauthorizedException(response)
 
     def _connect(self):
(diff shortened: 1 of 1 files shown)


drewaogle:
Before we revert, are we going to alter an expected behavior anywhere with this now?

gsaluja9:
> Before we revert, are we going to alter an expected behavior anywhere with this now?

no. This is how it has been. We will revisit the problems stated in the mac test, and try to repro the long running timeout scenario to think of a more comprehensive fix.

drewaogle:
Sounds good

Issue #475, opened 2024-09-06
nomenclature changes in SDK causing confusion.


gsaluja9:
SDK has classes with suffix model in pydantic derivatives.

Rename the following:
-------
- [x] QueryTypes -> DataModels
- [x] Model -> VideoDataModel
- [x] generate_save_query - > generate_add_query
- [x] execute_batch - > execute (and refactor out of PQ), Also change references to execute (from db.query)
- [x] db = create_connector() -> client = create_connector() (more apt meaning for what create_connector returns)
- [ ] #482

gsaluja9:
Also not call a generator unless it is a python generator ?

gsaluja9:
> Also not call a generator unless it is a python generator ?

@bovlb . I fear that this would need to be reworked separately. Even the way we've implemented Subscriptble is better served by a type in python called Sequence. I have read the docs [here](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence), but still not sure on the best course of action.

PR #481, opened 2024-09-12
Consistently named variables, and Disambiguation of terms used in SDK.

gsaluja9:
Addresses https://github.com/aperture-data/aperturedb-python/issues/475

### Naming alert.
Please look at the proposed naming scheme and the changes for consumers of aperturedb-python.
The major external impact of the change is the refactor of execute_batch out form ParallelQuery.

A Backward compatibility method is provided, but it'll log depreciation warnings.
The tests still work through that compatibility mechanism.

We would need to update demos and docs following this, though nothing should break, but deprecation warning .
This is an effort to generate leaner but clearer docs, and a more readable code in public repo


aperturedb/BlobNewestDataCSV.py
```diff
@@ -80,7 +80,7 @@ class BlobNewestDataCSV(CSVParser.CSVParser):
     ``` python
 
         data = ImageForceNewestDataCSV("/path/to/WorkingImageDataset.csv")
-        loader = ParallelLoader(db)
+        loader = ParallelLoader(client)
```
drewaogle, line 83: I'm worried `client` can be a confusing name as well. Do we know what the standard naming is that database libraries in python or in general use?

I feel like `conn` or `connection` is what I've seen a lot of.

gsaluja9: Let's stay on after standup and decide upon this.

aperturedb/Connector.py
```diff
@@ -496,16 +496,15 @@ def _renew_session(self):
                 time.sleep(1)
                 count += 1
 
-    def create_new_connection(self) -> Connector:
+    def clone(self) -> Connector:
```
luisremis, line 499: this is likely to break apps/demos/benchmarks. If this change is introduced, please make sure to make the changes on other repos.

drewaogle: or we can do what we did with the other things, make this a wrapper that throws depricated, and then make a log parser which finds depricated warnings? I think this might be useful in the future too.

gsaluja9: added the deprecated func as a wrapper with warning.

aperturedb/BlobNewestDataCSV.py
gsaluja9: So I brooded over the "memorable name" way of calling it, and I think it's not useful at this time. The vendors who do it are the big and popular companies, where the name of the product is synonymous with the brand. It'd be a good state to get to but I am focussed on keeping things simple to follow for now.

luisremis (approved):
thanks!

Issue #484, opened 2024-09-18
Long running Connector has no logic to reauthenticate.


gsaluja9:
Connector, which is called before a refresh token expires can issue a refresh token, which also extends the ttl of the refresh token.
A Connector that leads to invalid session response from Refresh Token does not ever kick off a authenticate to try to re create session.

Observed the following error log.

```
2024-09-18 08:12:16,581 : WARNING : aperturedb.Connector : 139877719430720 : 472 : Session expired while query was sent. Retrying... [semanticsearch.datasets.gcp.cloud.aperturedata.io:55555 as reader using TCP with SSL=True]
2024-09-18 08:12:16,606 : WARNING : aperturedb.Connector : 139877719430720 : 492 : [Attempt 1 of 3] Failed to refresh token.
Traceback (most recent call last):
  File "/opt/venv/lib/python3.10/site-packages/aperturedb/Connector.py", line 489, in _renew_session
    self._check_session_status()
  File "/opt/venv/lib/python3.10/site-packages/aperturedb/Connector.py", line 250, in _check_session_status
    self._refresh_token()
  File "/opt/venv/lib/python3.10/site-packages/aperturedb/Connector.py", line 265, in _refresh_token
    raise UnauthorizedException(response)
aperturedb.Connector.UnauthorizedException: [{'RefreshToken': {'info': 'Invalid Session', 'status': -1}}]
Stack (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 973, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 83, in _worker
    work_item.run()
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/opt/venv/lib/python3.10/site-packages/langchain_core/retrievers.py", line 194, in invoke
    return self.get_relevant_documents(
  File "/opt/venv/lib/python3.10/site-packages/langchain_core/_api/deprecation.py", line 148, in warning_emitting_wrapper
    return wrapped(*args, **kwargs)
  File "/opt/venv/lib/python3.10/site-packages/langchain_core/retrievers.py", line 316, in get_relevant_documents
[...]

PR #485, opened 2024-09-18
Re authenticate when refresh token fails.

gsaluja9:
fix for #484 

- [x] It recovers in subsequent calls to query. Make it kick in the current query.


luisremis (approved):
great, thanks!

PR #486, opened 2024-09-18
typer includes everything now .

gsaluja9:
no need for typer[all]

https://github.com/fastapi/typer/discussions/785


bovlb (approved):
LGTM

PR #487, opened 2024-09-19
Drop extra from typer

nicholsn:
As of typer 0.12.1 you no longer need to add the `typer[all]`, you can just use `typer`

See: https://typer.tiangolo.com/release-notes/#0121


nicholsn:
oops, someone beat me to it

PR #488, opened 2024-09-19
Numpy v2

bovlb:
This change, I believe fixes the problem with numpy version, so that we can use unpinned NumPy (e.g. 2.0.2) with newer (>=3.9.0) versions of Python, and pinned (<2) with previous versions.

I added `distributed` because it was flagged as a dependency for DASK in some versions.

See https://github.com/aperture-data/aperturedb-python/pull/445

We should probably let this one sit for a bit internally before packaging it into a version.


drewaogle (approved):
I verified this with containers with different version python.

PR #491, opened 2024-09-26
Fixed the loaders notebook in docs to work in colab.

gsaluja9:
Also addresses a bug in Images.py
Need to see why all polygons do not get loaded from annotations.


examples/CocoDataPyTorch.py
```diff
@@ -91,18 +95,38 @@ def generate_query(self, idx: int):
                 m = self.coco_detection.coco.annToMask(meta_info)
                 polygons = polygonFromMask(m)
                 adb_polygons = []
+                adb_boxes = []
                 for polygon in polygons:
                     adb_polygon = []
+                    left, top, right, bottom = 99999, 99999, 0, 0
```
drewaogle, line 101: I guess 99999 is ok for unset for coco?

gsaluja9: i've not seen images of that resolution in coco

drewaogle (changes requested):
Can we remove the outputs from loaders.ipyndb? it makes it very hard to review.

drewaogle (approved):
fixing the issue with diffing noteoboks is too hard here.

PR #493, opened 2024-09-30
switch declaration order in images

drewaogle:
When execution fails we get a spurious error of local variables not defined.


gsaluja9 (approved):
thanks.

PR #495, opened 2024-10-02
Make create_connection use dotenv and google colab secrets

bovlb:
This changes the behaviour of `create_connector` so that it can create a connection using a JSON object.  This object can come from one of three sources:
* Environment variable called `APERTUREDB_JSON`
* `.env` file entry  called `APERTUREDB_JSON`
* Google Colab secret called called `APERTUREDB_JSON`

### DIFF
--- a/aperturedb/CommonLibrary.py
+++ b/aperturedb/CommonLibrary.py
@@ -7,7 +7,7 @@ import importlib
 import math
 import os
 import sys
-from typing import Any, Callable, Optional, Tuple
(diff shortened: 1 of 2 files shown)


gsaluja9 (approved):
Thanks.
I see that walrus operator could be used at a bunch of other places in the SDK too.

Another thing is that in our CI testing, we do not use the create_connector at all. I remember that you mentioned we could do that for fixtures too. Would now be a good time to change that? I'd love to discuss the approach you had in mind.

PR #498, opened 2024-10-04
Handle colab non-notebook; auto-create config from colab secret

bovlb:
This PR has two changes:
* In the specific case of being in a Google Colab environment but not being in a notebook, we will not crash when attempting to access secrets.
* When using a connection from a Google Colab secret, we will automatically create and activate a config called "google_colab".


gsaluja9 (approved):
nice

PR #502, opened 2024-10-10
Add --from-json to adb config create

bovlb:
[ Recreation of https://github.com/aperture-data/aperturedb-python/pull/501 with lower-case branch name ]

Part of a set of changes to optimize quick start, etc.
This change adds a --from-json option to adb config create. If specified, then the normal connection parameters are ignored and instead we look for a JSON string in one of three places:

The environment variable APERTUREDB_JSON
An entry for APERTUREDB_JSON in a .env file
In interactive mode, the user will be prompted to enter the JSON string. This is treated like a password.

### DIFF
--- a/aperturedb/cli/configure.py
+++ b/aperturedb/cli/configure.py
@@ -131,7 +131,7 @@ def create(
     1) The environment variable APERTUREDB_JSON;
     2) An entry for APERTUREDB_JSON in a .env file;
     3) In interactive mode, the user will be prompted to enter the JSON string.  This will be treated as a password entry.
-    See https://docs.aperturedata.dev/Setup/client/notebooks for more information on JSON configurations.
+    See https://docs.aperturedata.dev/Setup/client/configuration for more information on JSON configurations.
     """
 
     db_host = host
--- a/aperturedb/cli/configure.py
+++ b/aperturedb/cli/configure.py
@@ -8,6 +8,7 @@ from typing_extensions import Annotated
 
 from aperturedb.cli.console import console
 from aperturedb.Configuration import Configuration
+from aperturedb.CommonLibrary import _create_configuration_from_json
 
 
 class ObjEncoder(JSONEncoder):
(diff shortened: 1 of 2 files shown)


aperturedb/cli/configure.py
```diff
@@ -149,28 +160,49 @@ def create(
...
+            except ImportError:
+                console.log("Unable to use the dotenv package")
+                pass
+        if json_str is None:
+            if interactive:
+                json_str = typer.prompt(
+                    "Enter JSON string", hide_input=True)
+        if json_str is None:
+            console.log(
+                "JSON string not found. Please set APERTUREDB_JSON environment variable create a .env file with APERTUREDB_JSON entry, or enter JSON in interactive mode")
```
vishakha041, line 179:
```suggestion
                "JSON string not found. Please set APERTUREDB_JSON environment variable create a .env file with APERTUREDB_JSON entry, or enter config parameters in interactive mode")
```

aperturedb/cli/configure.py
```diff
@@ -149,28 +160,49 @@ def create(
...
-        name=name,
-        host=db_host,
-        port=db_port,
-        username=db_username,
-        password=db_password,
-        use_ssl=db_use_ssl,
-        use_rest=db_use_rest
-    )
+    if from_json:
+        json_str = os.getenv("APERTUREDB_JSON")
```
vishakha041, line 164: So as of now, after I copy JSON from cloud , I will need to past it in a cell, fill my password and use that string ?

bovlb: You could, but my recommendation would be to prepare the JSON string in a more private editor.

vishakha041: OK - i was just thinking if it makes sense to prompt passwd and create this fully complete JSON in quick start or we explain that users have to paste a json with their passwd

vishakha041 (approved):
I didnt test it but I am guessing you hve.

aperturedb/cli/configure.py
bovlb: See https://github.com/aperture-data/Cookbook/pull/13 for the simplified Quickstart.

PR #505, opened 2024-10-10
Fix bug on check_status

gsaluja9:
This is the reason why there was no logging on non zero response on Create index when the index already exists scenario.

### DIFF
--- a/aperturedb/Connector.py
+++ b/aperturedb/Connector.py
@@ -43,6 +43,7 @@ from threading import Lock
 from types import SimpleNamespace
 from dataclasses import dataclass
 from aperturedb.Configuration import Configuration
+from aperturedb.types import CommandResponses
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


aperturedb/Connector.py
```diff
@@ -564,10 +564,10 @@ def last_query_ok(self):
...
 
     def check_status(self, json_res):
-
-        status = 0
-        if (isinstance(self.response, dict)):
-            if ("status" not in self.response):
+        # Default status is -2, which is an error, but not a server error.
+        status = -2
+        if (isinstance(json_res, dict)):
+            if ("status" not in json_res):
```
drewaogle, line 570: We should have a definition of statuses somewhere. -1 is server error, -2 is local, 0 is ok and 1+ are server responses?

gsaluja9: updating the doc string and types.

luisremis (approved):
thanks!

PR #510, opened 2024-10-22
Make name optional, add better error handling, and a little helpful output

bovlb:
In this morning's meeting, we wanted the user to be able to manage multiple cloud instances, using a name specified in the cloud UI to distinguish them.  In order to support this, we need the name argument to `adb config create` to be optional, which has a number of consequences.


gsaluja9 (approved):
This is beginning to have lot of paths, we should update the tests to get scenarios under coverage. Looks like what we discussed in the call yesterday. 

Merge it.

PR #511, opened 2024-10-22
Sparql query wrapper

bovlb:
Basic support for querying ApertureDB using SPARQL.

Suggested usage:
```
from aperturedb.sparql import SPARQL

sparql = SPARQL(client)
query = """..."""
df = sparql.to_dataframe(sparql.query(query))
```

Requires `pip install rdflib`. 

As part of adding tests, I created a test load of the Cookbook data, which is independently a good thing for us to have.  This required me to add clip and rdflib to the `[dev]` dependencies, which might be controversial.  I also adapted the interface of transformers so that they can take a client argument instead of calling `create_constructor()` directly.  This appears to be the first tests we have that use transformers.

### DIFF
--- a/aperturedb/SPARQL.py
+++ b/aperturedb/SPARQL.py
@@ -333,7 +333,7 @@ class SPARQL:
                 raise SPARQLError(
                     f"Unknown property or connection: {p}: valid connections are {self.connections.keys()}, valid properties are {self.properties.keys()}")
 
-        result, output, _ = execute_query(client, query, blobs)
+        result, output, _ = execute_query(self.client, query, blobs)
         solutions = set()  # process_bindings uses this to avoid yielding duplicate solutions
         yield from process_bindings(output)
 
--- a/aperturedb/SPARQL.py
+++ b/aperturedb/SPARQL.py
@@ -58,7 +58,8 @@ class SPARQL:
         self.knn_properties = set(self.namespaces["knn"] + p for p in [
                                   "similarTo", "set", "vector", "k_neighbors", "knn_first", "distance", "engine", "metric"])
 
-        self.logger.info(f"SPARQL loaded {len(self.connections)} connections and {len(self.properties)} properties")
+        self.logger.info(
+            f"SPARQL loaded {len(self.connections)} connections and {len(self.properties)} properties")
 
     def __del__(self):
         from rdflib.plugins.sparql import CUSTOM_EVALS
@@ -344,7 +345,8 @@ class SPARQL:
             self.output_response = output
             self.solutions = solutions
             self.triples = triples
-        self.logger.info(f"Evaluated BGP query with {len(triples)} triples, produced {len(commands)} commands, and {len(solutions)} solutions")
+        self.logger.info(
+            f"Evaluated BGP query with {len(triples)} triples, produced {len(commands)} commands, and {len(solutions)} solutions")
         return
 
     def _decode_descriptor(self, literal: "Literal") -> bytes:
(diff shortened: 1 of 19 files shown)


vishakha041:
It seems Nolan wants to see this wrapper whenever its public. He was curious when it came up.

aperturedb/SPARQL.py
```diff
@@ -0,0 +1,607 @@
...
+            command_body = list(query[i].values())[0]
+            result_body = list(output[i].values())[0]
+            entities = result_body.get("entities", [])
+
+            bindings2 = [(v, prop) for v, cb, prop in bindings
+                         if cb == command_body and not isinstance(v, BNode)]
+            assert all(isinstance(v, Variable)
+                       for v, _ in bindings2), bindings2
+            assert all(ctx[v] is None for v, _ in bindings2), bindings2
+            # print(
```
gsaluja9, line 244: We could logger debugs this .

aperturedb/SPARQL.py
```diff
@@ -0,0 +1,607 @@
...
+from aperturedb.CommonLibrary import create_connector
+from aperturedb.Utils import Utils
+
+
+class SPARQL:
+    def __init__(self, client=None, debug=False):
+        """SPARQL compatability layer for ApertureDB
+
+        Args:
+            client: ApertureDB client. If not uppliued, then `create_connector()` is used to create a new client.
```
gsaluja9, line 21:
```suggestion
            client: ApertureDB client. If not suppliued, then `create_connector()` is used to create a new client.
```

aperturedb/SPARQL.py
```diff
@@ -0,0 +1,607 @@
...
+                        command[prop] = o.toPython()
+                    else:
+                        add_binding(s2, p, o)
+            elif p == self.namespaces["c"] + "ANY":
+                add_is_connected_to(s, None, o)
+            else:
+                raise SPARQLError(
+                    f"Unknown property or connection: {p}: valid connections are {self.connections.keys()}, valid properties are {self.properties.keys()}")
+
+        output, _ = self._client.query(query, blobs)
```
gsaluja9, line 343:
```suggestion
        result, output, _ = aperturedb.CommonLibrary.execute_query(client, query, blobs)
```

aperturedb/SPARQL.py
bovlb: That specific print isn't very useful, but I'll add some logging.

pyproject.toml
```diff
@@ -67,7 +67,9 @@ dev = [
     "pre-commit",
     "pytest",
     "build",
-    "fuse-python ; platform_system == 'Linux'"
+    "fuse-python ; platform_system == 'Linux'",
+    "clip@git+https://github.com/openai/CLIP.git",
```
drewaogle, line 71: since this is for a test, should we target more than just their default branch - it doesn't look like it has any releases, unfortunately so maybe a specific commit?

test/conftest.py
```diff
@@ -26,21 +26,24 @@
...
+                    retry_interval_seconds=0)},
+                marks=pytest.mark.tcp),
+            pytest.param(
+                {"db": ConnectorRest(
+                    host=dbinfo.DB_REST_HOST,
+                    port=dbinfo.DB_REST_PORT,
+                    user=dbinfo.DB_USER,
+                    password=dbinfo.DB_PASSWORD,
+                    use_ssl=False)},
+                marks=pytest.mark.http)
```
drewaogle, line 46: this is literally world changing

gsaluja9 (approved):
neat! Thx for the conftest update

pyproject.toml
bovlb: pinned.

PR #516, opened 2024-10-31
Fix bug and add tests

bovlb:
I added some tests for `--from-json` and I found and fixed a bug with the new functionality.


gsaluja9 (approved):
Thanks! this is nice

PR #518, opened 2024-11-03
modify Images to support Frames as a subclass

drewaogle:
Modifies Images to support Frames, based on usage pattern from https://github.com/drewaogle/YOLOv4-OpenCV-CUDA-DNN/blob/main/VideoClips.ipynb which aimed to use `Images`, but on frame objects.

### DIFF
--- a/aperturedb/Images.py
+++ b/aperturedb/Images.py
@@ -14,6 +14,7 @@ from aperturedb import Utils
 from aperturedb.Entities import Entities
 from aperturedb.Constraints import Constraints
 from aperturedb.CommonLibrary import execute_query
+from aperturedb.Query import QueryBuilder, ObjectType,class_entity
 from ipywidgets import widgets
 from IPython.display import display, HTML
 import base64
@@ -97,7 +98,6 @@ def resolve(points: np.array, image_meta, operations) -> np.array:
     if image_meta["adb_image_width"] and image_meta["adb_image_height"]:
         image_meta_width = image_meta["adb_image_width"]
         image_meta_height = image_meta["adb_image_height"]
-        # print(f"resolve: {resolved}")
         for operation in operations:
             if operation["type"] == "resize":
                 x_ratio = operation["width"] / image_meta_width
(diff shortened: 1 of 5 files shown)


aperturedb/Images.py
```diff
@@ -139,6 +139,7 @@ class Images(Entities):
         client: The database connector, perhaps as returned by `CommonLibrary.create_connector`
     """
     db_object = "_Image"
+    query_object = "Image"
```
gsaluja9, line 142: This can be done in an alternate way, with Querybuilder.find_command(self.db_object, params) at every place you want to build FindCommand

aperturedb/Images.py
```diff
@@ -253,11 +254,12 @@ def __retrieve_batch(self, index):
         end = min(start + self.batch_size, len(self.images_ids))
 
         query = []
+        FindCommand = f"Find{self.query_object}"
```
gsaluja9, line 257: This variable name seems odd case, will not be needed if Querybuilder is used.

gsaluja9 (changes requested):
The variable named strikes odd and we can totally do away with creating a property .
The extension is fine in principle, but do consider asserting on frame specific params in image.

aperturedb/Images.py
drewaogle: Yes, I like that, but the goal was to minimize the changes. Do you want me to go and change everything to query builder?

gsaluja9: Yes . The problem to do it with different mechanisms can make a bigger to maintain surface area. Query builder is how it's mostly done in utilities DW , etc. 

The idea is to keep this query construction system self contained.

aperturedb/Images.py
drewaogle: changed to QueryBuilder

drewaogle:
changed to use `QueryBuilder` also made it so classes could use their `ObjectType` as the db_class rather than just a string

gsaluja9 (approved):
looks good.

aperturedb/Images.py
drewaogle: changed!

PR #526, opened 2024-11-21
Fix Images issues from Frames addition

drewaogle:
when I refactored `Images` I didn't check how I was referencing the FindCommand in the last pass and it was just looking for "Frame" or "Image" rather than "FindFrame" or "FindImage"

also the init for Frame was double passing responses.


gsaluja9 (approved):

gsaluja9 (approved):
merge it

PR #530, opened 2024-12-03
support protobuf5

drewaogle:


pyproject.toml
```diff
@@ -83,4 +83,4 @@ requires = ["setuptools>=61.0", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools.dynamic]
-version = {attr = "aperturedb.__version__"}
```
drewaogle, line 86: I don't know what added this - I did a `pip install -e` for testing. No idea if we want this or not.

gsaluja9: We do want this.

gsaluja9 (approved):
lgtm!

PR #531, opened 2024-12-04
add known hosts verification

drewaogle:
- Adds `VerifyType` to control how verification is done.
- sets `VerifyType.KNOWN_HOSTS` as the default action
- sets "*.aperturedata.io" as the known hosts.
- adds accessors for manipulating the known host list.
- adds `ADB_SSL_CA_FILE` as an environment variable usable for passing in a pem file for verification.

### DIFF
--- a/aperturedb/Connector.py
+++ b/aperturedb/Connector.py
@@ -118,8 +118,6 @@ class Connector(object):
 
 
     Some parameters are mainly of use when copying the object. `shared_data` and `authenticate` are designed for this.
-
-    
     """
 
     def __init__(self, host="localhost", port=55555,
--- a/aperturedb/Connector.py
+++ b/aperturedb/Connector.py
@@ -36,7 +36,7 @@ import json
 import ssl
 import logging
 import fnmatch
-from enum import Enum,auto
+from enum import Enum, auto
 from datetime import datetime, timedelta
 
 import keepalive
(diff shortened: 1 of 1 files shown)


drewaogle:
I've verified this works
- verifying connection to the cloud
- not verifying a connection to localhost
- failing to verify a self signed cert to localhost when localhost is added to know hosts.
- failing to verify a ca signed cert for localhost when the ca is missing from the ca store
- verifying with the ca signed cert for localhost after the ca is added.
- verifying with the ca signed cert for localhost with the ca cert passed in with ADB_SSL_CA_FILE

drewaogle:
If we add verification, I think we want to be able to know that we fail bad certs, as well as pass the expected cases, so if we do anything that were to change the operations, we would catch it.

I added the ADB_SSL_CA_FILE for hooking that up in tests, but adding those will be a lot of work.

ali-asadpoor:
@drewaogle Not sure if I understand the code, is this prevents client to accept self signed certs by default?

drewaogle:
> @drewaogle Not sure if I understand the code, is this prevents client to accept self signed certs by default?

By default it chooses enforcement by what host you are connecting to. If you are connecting to a *.aperturedata.io host, it will require a non-self-signed certificate.

drewaogle:
not usable

PR #536, opened 2024-12-06
Conditional add with transfomrers connecting to images

gsaluja9:


aperturedb/transformers/clip_pytorch_embeddings.py
```diff
@@ -30,11 +31,22 @@ def getitem(self, subscript):
 
         for ic in self._add_image_index:
             serialized = generate_embedding(x[1][ic])
+            # If the image already has an image_sha256, we use it.
+            image_sha256 = x[0][ic]["AddImage"].get("properties", {}).get(
+                "adb_image_sha256", None)
```
vishakha041, line 36: image_sha works since we control it but curious if it possible to use any unique constraint that user might have already specified in image csv?

vishakha041 (approved):

aperturedb/transformers/clip_pytorch_embeddings.py
gsaluja9: The thing is, these embedding transformers cannot assume anything since we do not enforce any schema . It does a best effort job of relying on adb_image_sha256 if available. We can discuss more options.

PR #548, opened 2025-01-14
option to get the output of transact as a json.

gsaluja9:


aperturedb/cli/transact.py
```diff
@@ -32,6 +32,19 @@ def load_fuse():
...
+def dump_as_raw_json(client: Connector, transaction: dict, **kwargs):
+    from aperturedb.CommonLibrary import execute_query
+
+    result, response, blobs = execute_query(
+        client=client,
+        query=transaction,
+        blobs=[])
+    print(json.dumps(response, indent=2))
+    for i, blob in enumerate(blobs):
+        print(f"len(blob[{i}]) = {len(blob[i])}")
```
bovlb, line 47: Unfortunately, this makes the output non-JSON.

bovlb (approved):
Looks good.  One comment. Did you find out where the trash was coming from?

aperturedb/cli/transact.py
gsaluja9: addressed it.

PR #550, opened 2025-01-16
Fix sparql blob order

bovlb:
I noticed that the blobs aren't been ordered according to the input list.


gsaluja9 (approved):
Looks good. I did not know that the blobs can be out of order.

luisremis:
I think blobs can be out of order only when using `group` or `group_by_source`. Otherwise, they should be ordered. If the they are not ordered, that's a bug on ApertureDB. I think it would break applications and tests.

PR #551, opened 2025-01-21
Check content length of image matches the content-length header

ali-asadpoor:
When downloading images or videos, sometimes content is not received completely and the `requests` library doesn't reflect any error. Comparing `Content-Length` header with the size of received content will tell us if content was downloaded completely or no.

### DIFF
--- a/aperturedb/Sources.py
+++ b/aperturedb/Sources.py
@@ -45,7 +45,7 @@ class Sources():
         retries = 0
         while True:
             imgdata = self.http_client.get(url)
-            if imgdata.ok:
+            if imgdata.ok and ("Content-Length" not in imgdata.headers or int(imgdata.headers["Content-Length"]) == imgdata.raw._fp_bytes_read):
                 imgbuffer = np.frombuffer(imgdata.content, dtype='uint8')
                 if not validator(imgbuffer):
                     logger.error(f"VALIDATION ERROR: {url}")


luisremis (approved):
thanks!

aperturedb/Sources.py
```diff
@@ -45,7 +45,7 @@ def load_from_http_url(self, url, validator):
         retries = 0
         while True:
             imgdata = self.http_client.get(url)
-            if imgdata.ok:
+            if imgdata.ok and ("Content-Length" not in imgdata.headers or int(imgdata.headers["Content-Length"]) == imgdata.raw._fp_bytes_read):
```
gsaluja9, line 48:
```suggestion
            if imgdata.ok and int(imgdata.headers.get("Content-Length", -1)) == imgdata.raw._fp_bytes_read:
```

gsaluja9: This could be simplified with dict.get . https://docs.python.org/3/library/stdtypes.html#dict.get

ali-asadpoor: @gsaluja9 Your simplification also changes the behavior, is that expected?

gsaluja9: No I did not mean to change. So we want to accept the response without content-length ? The default could be set to imgdata.raw._fp_bytes_read in that case.

ali-asadpoor: Yes.

gsaluja9 (approved):

PR #553, opened 2025-01-31
Add helpful errors for using a filtered DataFrame

drewaogle:
I have a dataframe I'm passing to ImageDataCSV, and it isn't happy for two reasons:

- CSVParser wants to use `RangeIndex.start` which doesn't exist on a filtered DataFrame.
- Then if you do reset_index, by default it then adds `index` which makes ImageDataCSV unhappy.

This PR simply tests for those conditions and gives a helpful error.
We may want to also add this to documentation. I don't think we want to attempt to correct inside the code because the supplied dataframes may be large.

### DIFF
--- a/aperturedb/CSVParser.py
+++ b/aperturedb/CSVParser.py
@@ -62,6 +62,10 @@ class CSVParser(Subscriptable):
                 self.df = pd.read_csv(filename)
             else:
                 self.df = df
+
+                if not hasattr( self.df.index, "start" ):
+                    raise ValueError("Supplied Dataframe's index missing 'start'" +
+                        "(Probably a filtered DataFrame - try reset_index )")
         else:
             # It'll impact the number of partitions, and memory usage.
             # TODO: tune this for the best performance.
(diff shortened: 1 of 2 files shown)


gsaluja9 (commented):
Hey @drewaogle . Is there a minimal use case that gets the run to this place. Can you add it somewhere.

I was under the impression that the CSV parser with df is only applicable to ingestion with dask. But this makes me to think that there are other use cases that lead to it.

aperturedb/CSVParser.py
```diff
@@ -62,6 +62,10 @@ def __init__(self,
                 self.df = pd.read_csv(filename)
             else:
                 self.df = df
+
+                if not hasattr(self.df.index, "start"):
+                    raise ValueError("Supplied Dataframe's index missing 'start'" +
+                                     "(Probably a filtered DataFrame - try reset_index )")
```
bovlb, line 68:
```suggestion
                                     "(Probably a filtered DataFrame - try reset_index(drop=True) )")
```

bovlb (approved):
Suggestion to avoid adding index

PR #556, opened 2025-02-13
Remove progressbar.

gsaluja9:


luisremis (approved):
I think having ETA is a very important psyco-emotional feature (?). I will miss it.

gsaluja9:
> I think having ETA is a very important psyco-emotional feature (?). I will miss it.

ETA is still around.

PR #563, opened 2025-02-26
Allow notbook_dir to be configured

ali-asadpoor:


ali-asadpoor:
Tested locally and confirmed works as expected.

gsaluja9 (approved):
looks good.

PR #568, opened 2025-03-10
Simple Tokenizer to aid users adding configurations

drewaogle:


aperturedb/Connector.py
```diff
@@ -113,7 +114,8 @@ def __init__(self, host="localhost", port=55555,
                  use_keepalive=True,
                  retry_interval_seconds=1,
                  retry_max_attempts=3,
-                 config: Optional[Configuration] = None):
+                 config: Optional[Configuration] = None,
+                 encoded: Optional[str] = None):
```
gsaluja9, line 118: Can we just call this token?

aperturedb/cli/configure.py
```diff
@@ -124,7 +124,9 @@ def create(
             help="Interactive mode")] = True,
         overwrite: Annotated[bool, typer.Option(
             help="overwrite existing configuration")] = False,
-        from_json: Annotated[bool, typer.Option(help="create config from a JSON string")] = False):
+        from_json: Annotated[bool, typer.Option(
+            help="create config from a JSON string")] = False,
+        from_encoded: Annotated[bool, typer.Option(help="create config from an encoded string")] = False):
```
gsaluja9, line 129: from_token

gsaluja9 (changes requested):
This also needs to add a case in CommonLibrary.create_connector to check for a presence of ADB_TOKEN and create it  even if no other config is available.

aperturedb/Connector.py
drewaogle: we absolutely can. I think I called it "encoded" just because "token" is already overloaded in our codebase with the token generation for account access, etc - 

if we make sure people know this is an "connect token" or something hopefully it wouldn't cause issues.

drewaogle:
> This also needs to add a case in CommonLibrary.create_connector to check for a presence of ADB_TOKEN and create it even if no other config is available.

Yes, for sure.

I've also made a greater compression version of this which adds options to `adb` for creating tokens.

I think at this point if everything else looks good we should decide what we want to target as a token length, and features.

aperturedb/cli/configure.py
drewaogle: changed to `key` as we discussed today

aperturedb/cli/configure.py
```diff
@@ -124,7 +123,9 @@ def create(
...
```
gsaluja9, line 128: This still refers it as "from_encoded"

gsaluja9 (commented):
Do we also want to add an check for ADB_KEY in os.environment in create_connector.
That would be super useful for our integrations to work seamlessly.

**Examples that would benefit:**
https://github.com/langchain-ai/langchain/blob/8a69de5c2464f7a451e4a82606331e8d1124a1a3/libs/community/langchain_community/vectorstores/aperturedb.py#L98

https://github.com/aperture-data/llama_index/blob/458443732e061822a824a25cc10cd0d50e0f27e9/llama-index-integrations/vector_stores/llama-index-vector-stores-ApertureDB/llama_index/vector_stores/ApertureDB/base.py#L105

aperturedb/cli/configure.py
drewaogle: fixed.

gsaluja9 (approved):
thanks

PR #569, opened 2025-03-13
Change defaults to HNSW and Cosine similarity

gsaluja9:
Use CS in CLIP embeddings, as it works the best.

### DIFF
--- a/aperturedb/Utils.py
+++ b/aperturedb/Utils.py
@@ -473,7 +473,8 @@ class Utils(object):
 
         return total_connections
 
-    def add_descriptorset(self, name: str, dim: int, metric="L2", engine="FaissFlat",
+    def add_descriptorset(self, name: str, dim: int, metric=["L2", "CS", "IP"],
+                          engine=["Flat", "HNSW"],
                           properties: Optional[Dict] = None) -> bool:
         """
         Add a descriptor set to the database.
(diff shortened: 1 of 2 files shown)


aperturedb/Utils.py
```diff
@@ -473,7 +473,8 @@ def count_connections(self, connections_class, constraints=None) -> int:
 
         return total_connections
 
-    def add_descriptorset(self, name: str, dim: int, metric="L2", engine="FaissFlat",
+    def add_descriptorset(self, name: str, dim: int, metric=["L2", "CS", "IP"],
```
luisremis, line 476: I don't think we should do this by default. Usually, in real use cases, you use a single metric. The embeddings are generated in such a way that a specific metric will work best. 

This would, by default, replicate the data in 6 different faiss indexes.

gsaluja9: fair enough. I will set defaults as : "HNSW", "CS"

aperturedb/Utils.py
```diff
@@ -473,7 +473,8 @@ def count_connections(self, connections_class, constraints=None) -> int:
...
+                          engine=["Flat", "HNSW"],
```
gsaluja9, line 477:
```suggestion
    def add_descriptorset(self, name: str, dim: int, metric=["CS"],
                          engine=["HNSW"],
```

luisremis (approved):

PR #570, opened 2025-03-13
Fix stacktrace invocation making import impossible in windows

drewaogle:
user on ApertureDB slack reported issue.

https://docs.python.org/3/library/faulthandler.html#dumping-the-traceback-on-a-user-signal

`register` is not available on windows; it will simply throw an error if you try to sue it.

### DIFF
--- a/aperturedb/__init__.py
+++ b/aperturedb/__init__.py
@@ -8,15 +8,21 @@ from string import Template
 import platform
 import faulthandler
 import signal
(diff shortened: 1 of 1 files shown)


gsaluja9 (approved):
looks good. But I'd advice caution and very least running the tests before leading the unblocked user on with the library. I am still very unsure of using the SDK on windows.

PR #576, opened 2025-04-02
Improve run time by running http and non http tests concurrently

gsaluja9:
This run_test completed in 40 mins.


luisremis (dismissed):
thanks!

PR #577, opened 2025-04-02
remove grpcio-status pin

drewaogle:
it appears that grpcio-status was pinned to aid in installing `google-cloude-bigquery`

however now:
```
python3 -m venv ./bqtest
./bqtest/bin/activate
pip install -e .
pip install google-cloud-bigquery
```

works fine.

Results are:
```
(bq_test) drew@Dunstan:/code/aperture/aperturedb-python$ pip list | grep grpc
grpcio                   1.71.0
grpcio-status            1.71.0
```

versus old:
```
(adb_testing) drew@Dunstan:~$ pip list | grep grpc
grpcio                    1.63.0
grpcio-status             1.48.2
```

so it looks like a newer gprc can be used as well.


gsaluja9 (approved):
thanks!

PR #581, opened 2025-04-08
Updating to the prescribed version tags.

gsaluja9:


gsaluja9:
> @gsaluja9 Can you please take a look at https://github.com/aperture-data/aperturedb-python/blob/develop/test/.env too?
done.

ali-asadpoor (approved):
Thanks!

PR #589, opened 2025-05-08
Add 'remove' for configurations

drewaogle:
* Removes 1 configuration
* can operate at global or local
* refuses to remove active config unless given options.
* can either take a config to set active, or can just assign one.

### DIFF
--- a/aperturedb/cli/configure.py
+++ b/aperturedb/cli/configure.py
@@ -39,7 +39,7 @@ def _config_file_path(as_global: bool) -> Path:
     return config_path
 
 
-def _write_config(config_path:Path, config: dict):
+def _write_config(config_path: Path, config: dict):
     with open(config_path.as_posix(), "w") as config_file:
         config_file.write(json.dumps(config, indent=2, cls=ObjEncoder))
 
--- a/aperturedb/cli/configure.py
+++ b/aperturedb/cli/configure.py
@@ -126,7 +126,6 @@ def ls(log_to_console: bool = True):
 
     return all_configs
 
-
 @app.command()
 def create(
         name: Annotated[Optional[str], typer.Argument(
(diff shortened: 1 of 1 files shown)


aperturedb/cli/configure.py
```diff
@@ -270,6 +270,61 @@ def activate(
         config_file.write(json.dumps(configs, indent=2, cls=ObjEncoder))
 
 
+@app.command()
+def remove(
+        name: Annotated[Optional[str], typer.Argument(
+            help="Name of this configuration to remove")],
+        remove_if_active: Annotated[bool, typer.Option(
```
bovlb, line 277: I'm not sure these parameters are worth the complexity.
If you want to activate another config first, we already have an action for that.
Automatically activating an arbitrary config seems like a recipe for confusion.

aperturedb/cli/configure.py
```diff
@@ -270,6 +270,61 @@ def activate(
...
+            raise typer.Exit(code=2)
+
+    del configs[name]
+    if change_active:
+        if new_active:
+            ac = new_active
+        else:
+            ac = next(iter(configs))
+    configs["active"] = ac
+    with open(config_path.as_posix(), "w") as config_file:
```
bovlb, line 324: This should be a helper method

bovlb (dismissed):
Looks good. The original motivating example was that we wanted to create an ephemeral config for one task, and then restore the previous set up, which this doesn't quite provide.

aperturedb/cli/configure.py
drewaogle: this isn't activating another config first - by default it will fail to delete an active config.
the `remove-if-active` is kind of the `rm -rf` of 'just get rid of it'

aperturedb/cli/configure.py
drewaogle: moved to helper

gsaluja9 (dismissed):
looks good.

PR #594, opened 2025-06-17
Adds a class that can iterate over Croissant datasets and generate queries to persist them to ApertureDB.

gsaluja9:
This can be invoked as follows:
For a dry run (all the generated queries in files in cwd):
```
adb ingest from-croissant https://huggingface.co/api/datasets/FreedomIntelligence/PubMedVision/croissant --sample-count 1 --debug
adb ingest from-croissant https://huggingface.co/api/datasets/Rapidata/text-2-video-human-preferences-veo3/croissant --sample-count 1 --debug
adb ingest from-croissant https://huggingface.co/api/datasets/MathLLMs/MathVision/croissant --sample-count 1 --debug
```

Without the trailing --debug to actually run these queries on ApertureDB.

TODO:
- [ ] URL resolution
- [ ] Flattening of nested JSON into connected entities.
- [ ] Avoid intermediate parse into a Dataframe.

### DIFF
--- a/aperturedb/CommonLibrary.py
+++ b/aperturedb/CommonLibrary.py
@@ -315,7 +315,7 @@ def execute_query(client: Connector, query: Commands,
                     warn_list.append(wr)
         if len(warn_list) != 0:
             logger.warning(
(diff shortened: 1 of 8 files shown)


luisremis (approved):
🚀

PR #595, opened 2025-06-19
Croissant data enrichment

gsaluja9:
1. example of getting a URL ingested as a blob.
2. option of flattening a json into connected entities (1 level)
3. Has no intermediate conversion into a Dataframe


luisremis (approved):
thanks!

PR #599, opened 2025-07-10
Small fixes to CSV df and Configuration token

drewaogle:
CSVLoader actually wants an RangeIndex, since everything wants to use `start` and only RangeIndex has it. It's better to catch it earlier.

I missed passing token to Configuration when loading from the config file, so if you save a aperturedb key, it will lose the token when it loads.

### DIFF
--- a/aperturedb/CSVParser.py
+++ b/aperturedb/CSVParser.py
@@ -76,6 +76,11 @@ class CSVParser(Subscriptable):
                 self.filename,
                 blocksize=blocksize)
 
+        # we expect the df index to have 'start', which means RangeIndex.
+        # most users don't supply their own df, so this is mostly a sanity check
+        # for when an advanced user has done filtering and have a IntervalIndex.
+        if not isinstance(self.df.index,pd.RangeIndex):
+            raise Exception(f"CSVParser requires a RangeIndex. the supplied DataFrame has a type(self.df.index) index.")
         # len for dask dataframe needs a client.
         if not self.use_dask and len(self.df) == 0:
             logger.error("Dataframe empty. Is the CSV file ok?")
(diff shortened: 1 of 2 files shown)


gsaluja9 (dismissed):
lgtm

aperturedb/CSVParser.py
```diff
@@ -62,7 +62,16 @@ def __init__(self,
                 self.df = pd.read_csv(filename)
             else:
                 self.df = df
+
+            # we expect the df index to have 'start', which means RangeIndex.
+            # most users don't supply their own df, so this is mostly a sanity check
+            # for when an advanced user has done filtering and have a IntervalIndex.
+            if not isinstance(self.df.index, pd.RangeIndex):
+                raise Exception(
```
bovlb, line 70:
```suggestion
                raise TypeError(
```

aperturedb/CSVParser.py
```diff
@@ -62,7 +62,16 @@ def __init__(self,
...
+                    f"CSVParser requires a RangeIndex. the supplied DataFrame has a {type(self.df.index)} index.")
         else:
+            if df is not None:
+                raise Exception("Dask requires a csv passed as a filename")
```
bovlb, line 74:
```suggestion
                raise ValueError("Dask requires a csv passed as a filename")
```

aperturedb/cli/configure.py
```diff
@@ -75,7 +75,8 @@ def get_configurations(file: str):
                 username=config["username"],
                 password=config["password"],
                 use_rest=config["use_rest"],
-                use_ssl=config["use_ssl"])
+                use_ssl=config["use_ssl"],
+                token=config["token"] if "token" in config else None)
```
bovlb, line 79:
```suggestion
                token=config.get("token"),
            )
```

bovlb (approved):
LGTM

PR #601, opened 2025-08-04
Update support for GetSchema

luisremis:
Adds support for:
https://github.com/aperture-data/athena/issues/757#issuecomment-2867364234 

Unblocks: https://github.com/aperture-data/athena/pull/1726


gsaluja9 (approved):
Thanks !

PR #614, opened 2025-09-05
Connector config unification

drewaogle:
Simplify parsing of options to Connector ( fixes a bug I had with passing in a key not setting use_keepalive )
adds documentation to say how it works.


gsaluja9 (approved):
lgtm

Issue #533, opened 2024-12-05
Connector must validate the SSL certificate and self signed should be rejected by default.


gsaluja9:
This can be over ridden by a parameter.
Affected artifacts would be 
- adb config *
- documentation 
- Some demos

Anywhere in user code where create_connector() is called, or Connector() is instantiated.

vishakha041:
@gsaluja9  How does it impact the users that are just curious developers trying or people within companies but without approval yet?
We have two scenarios - one docker and one cloud that people use to try. 

@drewaogle  is this required by SOC2 ?

drewaogle:
@vishakha041 no, this is not required by SOC2.

drewaogle:
@vishakha041 but more to the point, I think this is a tricky security situation.

We don't want to make examples that say verify=false, as that will cause bad security, but for people just trying it out, if they don't use the cloud, they will get a confusing error.

One solution is to make the server not do TLS by default, but this isn't great, because you're decreasing security to .. increase security?

gsaluja9:
> [@vishakha041](https://github.com/vishakha041) but more to the point, I think this is a tricky security situation.
> 
> We don't want to make examples that say verify=false, as that will cause bad security, but for people just trying it out, if they don't use the cloud, they will get a confusing error.
> 
> One solution is to make the server not do TLS by default, but this isn't great, because you're decreasing security to .. increase security?

I also gave this a considerable thought. How about we introduce some params  in config, and offer 3 tiered solution in the SDK. Right now, we have 2, which is use_ssl = True | False

use_ssl = same meaning
if use_ssl is True:
   use_ssl_hostname_verification = boolean to do a full certificate and hostname check.
   ca_cert : path to pre shared ca certificate.

luisremis:
Done

PR #616, opened 2025-09-24
Verify hostname and issuer certificate by default.

gsaluja9:
For addressing #533 

Presently the hostname verification  not enforced in the Connector.
This was an aid to ease deployment and local development.

This PR adds the hostname verification back into the Connector* .
This will obviously create an extra step to jump for setting up for local development.

The devcontainer introduced as a part of this PR provides a preconfigured environment using a issued certificate.
TODO:
- [x] Enable it in tests
- [x] Provide more info in setup local in [docs](https://docs.aperturedata.io/Setup/server/Local#run-aperturedb-along-with-webui-using-docker-compose) (It uses self signed certificate) . [PR](https://github.com/aperture-data/docs/pull/284)
- [x] The [one liner](https://docs.aperturedata.io/Setup/server/Local#non-mac-systems) docker command would need to be changed to not use_ssl.
- [x] More docs to add transportation and use of custom CA would need to be created.
- [x] Key creation needs and versioning logic needs an update. Thanks @drewaogle for the update, and explanation.

### DIFF
--- a/aperturedb/Connector.py
+++ b/aperturedb/Connector.py
@@ -74,6 +74,8 @@ RENEW_SESSION_RETRY_INTERVAL_SEC = 1
 STATUS_OK = 0
 STATUS_ERROR_DEFAULT = -2
 
+SETUP_URL = "https://docs.aperturedata.dev/Setup/server/Local#run-aperturedb-along-with-webui-using-docker-compose"
+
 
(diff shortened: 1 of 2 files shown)


drewaogle (approved):
lgtm

drewaogle (approved):
lgtm

PR #624, opened 2025-10-01
using default context for preconfigured settings + better logging.

gsaluja9:


aperturedb/Connector.py
```diff
@@ -376,7 +376,7 @@ def _connect(self):
             if self.use_ssl:
 
                 # Server is ok with SSL, we switch over SSL.
-                self.context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+                self.context = ssl.create_default_context()
```
drewaogle, line 379: what is `create_default_context` supposed to be? I only ask because usually in security contexts its better to be explicit about what you are configuring if possible.

aperturedb/Connector.py
```diff
@@ -396,7 +396,8 @@ def _connect(self):
                         f"The host name must match the certificate: {self.host} " + os.linesep + \
```
bovlb, line 396: * Should use `logger.exceeption` here to automatically report exception and traceback
* These three calls to logger should be one
* This call to assert might be better as `raise ssl.SSLError('...') from e`

aperturedb/Connector.py
```diff
@@ -396,7 +396,8 @@ def _connect(self):
...
                         f"You can use the ca_cert parameter to specify a custom CA certificate " + os.linesep + \
                         f"Refer to the documentation for more information: {SETUP_URL}" + os.linesep + \
-                        f"Alternatively, SSL can be disabled by setting use_ssl=False (not recommended)"
+                        f"Alternatively, SSL can be disabled by setting use_ssl=False (not recommended)" + os.linesep + \
+                        f"{e=}"
                 except ssl.SSLError as e:
                     logger.error(f"Error wrapping socket: {e}")
                     self.conn.close()
```
bovlb, line 403: I don't think it's necessary to close the connection here, given the `BaseException` catch below.

PR #626, opened 2025-10-02
Release 0.4.51

gsaluja9:
- brings in compatibility with newer aperturedb get schema response.
- Enforces ssl host name verification by default.
- Misc. bug fixes.


bovlb (approved):
I hope this SSL thing isn't going to bite us. I'd rather we had more time to play with that before pushing it out.

PR #653, opened 2026-04-21
fix: cleanup previous runner state properly to avoid permission denied errors in self-hosted runners

ad-claw000:
Closes #N/A

This PR fixes the failing GitHub actions. The self-hosted runners are failing to run the `actions/checkout` step because the `test/aperturedb/logs` and `db*` folders created by previous docker runs are owned by root and cannot be removed by git clean. Adding a pre-checkout docker cleanup step solves this.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -12,6 +12,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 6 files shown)


luisremis:
solved on `release-0.4.56` branch https://github.com/aperture-data/aperturedb-python/pull/655

PR #680, opened 2026-05-19
fix(logging): combine exception and warning log messages

### LINKED ISSUE #408: Combind logger exception/warning
in Connector.py there are logger calls in exception handlers that log the exception separately from the message, this should almost certainly be a single call?

ad-claw000:
Closes #408

This replaces consecutive `logger.exception` and `logger.warning` calls with a single `logger.warning(..., exc_info=True)` where appropriate, to avoid splitting the exception traceback from the warning message.

Changes:
- In `Connector.py`, SSL and OS errors are now logged as warnings containing the full traceback context.
- In `ParallelQuery.py`, batch execution errors are logged similarly.

Tested locally via syntax checks.

### DIFF
--- a/aperturedb/Connector.py
+++ b/aperturedb/Connector.py
@@ -500,24 +500,20 @@ class Connector(object):
                 if tries != 0 or (self.last_query_timestamp is not None and
                                   (now - self.last_query_timestamp) <
                                   self.query_connection_error_suppression_delta):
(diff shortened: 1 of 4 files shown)


ad-claw000:
Closing in favor of #684 which cleanly isolates the logging fixes from other PRs.

PR #689, opened 2026-05-19
fix: remove blocking pypi version check

### LINKED ISSUE #182: Package version checks requires internet access to hit pypi
This is a blocking call:
https://github.com/aperture-data/aperturedb-python/blob/develop/aperturedb/__init__.py#L34

When no internet access, this will block for about a minute before failing. We should not need package to connect to internet in the future.

ad-claw000:
Closes #182

Removed the blocking internet check for the latest version on PyPI during import. This resolves the issue where importing the library without internet access would hang for about a minute before failing.

### DIFF
--- a/aperturedb/Utils.py
+++ b/aperturedb/Utils.py
@@ -173,57 +173,58 @@ class Utils(object):
         entities = r['entities']['classes']
         connections = r['connections']['classes']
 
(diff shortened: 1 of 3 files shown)


luisremis:
fixed by https://github.com/aperture-data/aperturedb-python/pull/652

PR #696, opened 2026-05-19
feat: Support dynamic batchsize based on query size

### LINKED ISSUE #328: ParallelQuery batchsize cannot adjust for query size
Generally it is fastest to make the batchsize as large as possible, and the effective limit is on the total byte size of the encoded query.  It would be useful if we could make the batchsize adjust automatically to keep the encoded query under a specific byte limit.

ad-claw000:
## Summary
Allows ParallelQuery and ParallelLoader to dynamically adjust batch size based on the estimated bytes of the query + blobs to prevent large payload rejection and to allow user to limit memory consumption per batch. 

## Verification
Tests using `ParallelQuery` and `ParallelLoader` complete correctly and dynamic batching logs show when batch limits are hit. The feature is tested against large blobs estimating query bounds correctly. 

Fixes aperture-data/aperturedb-python#328.

### DIFF
--- a/aperturedb/ParallelLoader.py
+++ b/aperturedb/ParallelLoader.py
@@ -164,7 +164,7 @@ class ParallelLoader(ParallelQuery.ParallelQuery):
                         logger.warning(
                             f"Failed to create index for {connection_class}.{property_name}")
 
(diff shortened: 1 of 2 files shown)


luisremis:
duplicated with #669

PR #704, opened 2026-05-20
fix(cli): reduce traceback length by disabling Typer pretty exceptions

### LINKED ISSUE #467: SDK trial on Mac

ad-claw000:
## Summary
* Disabled `pretty_exceptions_enable` on Typer apps across the CLI commands to revert to standard Python tracebacks.
* This addresses the issue where the rich error stack traces were excessively long and mistaken for `IPython` causing slowdowns.

## Verification
* Verified that a deliberate exception raised inside a Typer command produces a standard Python traceback instead of a rich Typer stack trace.
* Checked `adb config ls` command execution still behaves correctly.

Fixes aperture-data/aperturedb-python#467.


ad-claw000 (commented):
Applied autopep8 formatting to the new log messages to fix the pre-commit CI failure.

ad-claw000 (commented):
Fixed the pre-commit CI failure by applying autopep8 formatting to the new log messages. Ready for review!

PR #731, opened 2026-05-21
ci: ignore docker push failure for tests image

ad-claw000:
The build on `develop` is failing because it attempts to push `aperturedb-python-tests` to Docker Hub, but the repository either does not exist or the provided access token lacks sufficient scope.

This PR ignores the `docker push` failure for the tests image, which unblocks the CI pipeline. The underlying permissions or repository creation on Docker Hub should be addressed to enable caching as intended by the previous PR.

Ref: https://github.com/aperture-data/aperturedb-python/actions/runs/26204770897/job/77108340433


luisremis:
token fixed

PR #737, opened 2026-05-21
Fix: ParallelQuery batchsize cannot adjust for query size

### LINKED ISSUE #328: ParallelQuery batchsize cannot adjust for query size

ad-claw000:
Summary
Added `max_batchsize_bytes` to `ParallelQuery` and `ParallelLoader` to dynamically adjust the number of elements in a batch to prevent the encoded query from exceeding the limit.

Verification
Ran `pytest` tests.

Fixes aperture-data/aperturedb-python#328


luisremis:
covered by #669

PR #740, opened 2026-05-23
Closes #739: Skip slow and external network tests in PR CI

### LINKED ISSUE #739: Investigate and reduce CI execution times for PR workflow
The CI times for the `pr.yaml` workflow in `aperturedb-python` are quite long (e.g., `run_test_conda_gpu` taking over 1 hour when it completes, and `run_test` jobs varying from 30m to over an hour). This negatively impacts the developer feedback loop on PRs. 

We can explore several approaches to reduce these execution times:

### 1. Split Pytest Execution (Test Sharding/Parallelization)
The `run_test.sh` script currently runs all `pytest` markers sequentially within the Docker container for a given protocol.
* **Approach**: We can split the test suite into parallel chunks using `pytest-split` or `pytest-xdist`, or by sharding across different GitHub Actions jobs based on test sub-directories (e.g., separating `test_Datawizard.py` and `test_EntityLoader.py` into different jobs).
* **Pros**: Direct reduction in wall-clock time for the test execution phase.
* **Cons**: Requires managing test coverage aggregation across multiple shards.

### 2. Cache Docker Layers and Pip Dependencies
The workflow builds multiple Docker images (e.g., `docker/tests/Dockerfile` and `docker/pytorch-gpu/Dockerfile`). Re-running `pip install -e ".[dev]"` and cloning dependencies like `openai/CLIP` takes several minutes.
* **Approach**: Utilize GitHub Actions Docker Layer Caching (`docker/setup-buildx-action` with `actions/cache`) or export intermediate Docker layers to the GitHub Container Registry (ghcr.io) and pull them as cache sources (`--cache-from`).
* **Pros**: Speeds up the Docker build phase of `run_test` and `run_test_conda_gpu`.
* **Cons**: Caching large Docker images can sometimes hit GitHub Actions storage limits or incur cache save/restore overhead.

### 3. Separate "Slow" / "External" Tests into a Nightly Workflow
There are multiple tests marked with `slow`, `external_network`, and `remote_credentials`. Downloading external images and setting up large Dask dataframes is time-consuming.
* **Approach**: Deselect `@pytest.mark.slow` and `external_network` from the PR workflow by passing `-m "not slow and not external_network"` and move these longer end-to-end integration tests to a nightly cron workflow.
* **Pros**: Substantial reduction in PR CI time.
* **Cons**: PRs might inadvertently break external network or slow integration paths without immediate feedback, delaying the discovery of the regression until the nightly run.

ad-claw000:
Closes #739

This PR addresses the slow CI execution times in the PR workflow by implementing Option 3 from the issue: **Separating 'Slow' / 'External' Tests from PR workflows**.

### Why Option 3 over the others?

1. **Option 1 (Sharding pytest)**: Pytest sharding using `pytest-split` or `pytest-xdist` would reduce wall-clock time, but it introduces the complexity of aggregating coverage across jobs. It also runs the risk of race conditions if concurrent test processes mutate the same ApertureDB database state.
2. **Option 2 (Caching)**: We already utilize `DOCKER_BUILDKIT: 1` and layer caching via GitHub Container Registry (see `pr.yaml`). While we could attempt to cache Pip more granularly, the actual Pytest execution phase remains the largest bottleneck.
3. **Option 3 (Separating slow tests)**: By simply setting an environment variable (`SKIP_SLOW_TESTS=true`) exclusively in the PR workflow, we gain immediate and significant reduction in CI completion times. The tests marked with `@pytest.mark.slow` or `@pytest.mark.external_network` will be skipped in PR feedback loops, but will continue running in the daily `dependencies.yml` runs and `develop.yml` / `main.yml` workflows to ensure no coverage is lost globally.

### Implementation details

- Modified `pr.yaml` to pass `SKIP_SLOW_TESTS=true`
- Updated `run_test_container.sh` to forward this environment variable directly to the `docker run` commands executing tests
- Updated `run_test.sh` to inject `and not slow and not external_network` into the pytest `$FILTER` query when the flag is true

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,56 @@
+name: nightly
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-test:
+    runs-on:
+    - benchmark
+
+    steps:
+
+    - uses: actions/checkout@v3
+
+    - name: Login to DockerHub
+      uses: docker/login-action@v2
+      with:
+        username: ${{ secrets.DOCKER_USER }}
+        password: ${{ secrets.DOCKER_PASS }}
+
+    - name: Login to Google Cloud
+      uses: google-github-actions/setup-gcloud@v0
+      with:
+        service_account_key: ${{ secrets.GCP_SERVICE_ACCOUNT_KEY }}
+        project_id:          ${{ secrets.GCP_SERVICE_ACCOUNT_PROJECT_ID }}
+        export_default_credentials: true
+
+    - name: Build and Run Tests
+      env:
+        # Run both protocols inside a single job. run_test_container.sh starts
+        # the http and non_http stacks in parallel, so wall-clock time is
+        # roughly max(http, non_http) instead of the previous matrix approach
+        # which duplicated every docker image build (notebook deps / notebook
+        # / tests / coverage) on a second runner.
+        TEST_PROTOCOL: both
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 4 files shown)


luisremis (changes requested):
you need to add the nightly workflow, which runs all the tests, as part of this pr.

ad-claw000:
Added `nightly.yml` workflow to run all tests on a schedule, as requested. See commit 88f1e44.

